PR #2, opened 2012-10-26
Fixing fonts problem with PDF version 1.7, created by ABBY

gustavo-moreira:
PDFBox was throwing a NullPointerException when the PDF file did not provide the COSName.TYPE information. In a PDF created by ABBY (PDF version 1.7), this information is missing, but the Font SubType and the BaseFont still there, so the content can still be extracted correctly.

In PDFontFactory was just changed the Font information retrieve, a implemented a fallback for the missing information.


jahewson:

PR #4, opened 2013-05-01
Fix closure of exported resource

jmkgreen:
This fixes https://issues.apache.org/jira/browse/PDFBOX-1586 although there will be an underlying problem which has not been inspected. Solves problem of crashes when saving split fragments from a large document,


jahewson:

PR #7, opened 2013-06-13
Update DocumentEncryption.java

scotty1024:
Altered to not assign value to argument "password"
Added final declarations to many copied values.
Altered id to byte[] as it was never used as a COSString but rather as the getBytes() from that.
Added "passwordBytes" to hold reference to password.getBytes()
Renamed "o" to "ownerPassword"
Renamed "u" to "userPassword"


jahewson:

THausherr:
I didn't use that one although I'm aware of such issues from our Sonar reports... it just makes the code more complex. We don't act on all complaints from Sonar, but we still do a lot... if you have changes related to code quality, you can submit them here: https://issues.apache.org/jira/browse/PDFBOX-2852 . When I started to look at the reports (Dec 2014) we were below 90% rule compliance, now we're at 92.5%, which puts us at the top of the 2nd third of all Apache projects:
https://analysis.apache.org/dashboard/?did=5

PR #8, opened 2013-06-13
Update InstructionSequenceBuilder.java

scotty1024:
Altered to not change passed argument "token".
Added comment that in Java 7 this wrapper method is no longer required.


jahewson:

THausherr:
change done. Better late than never :-)

PR #15, opened 2015-10-08
PDFAcroForm flatten() ClassCastException

jasonmackin:
When trying to flatten a pdf, I was was receiving a ClassCastException.   field.getWidgets() can return both PDTerminalField and PDNonTerminalField. The for loop casts it to a PDTerminalField.

Added a simple if statement checking the instanceof to make sure it was a PDTerminalField before casting and it seems to have fixed the problem.


benmccann:

PR #17, opened 2015-10-30
Mark transparency groups

kalaspuffar:
We try to read text from PDF files but some of the files include extra data that is never shown. These segments are usually grouped in transparency groups. So for us this function to flag a marked content as a transparency group is quite useful.

If there is a way to do this please tell me or if there is a better way to remove text that isn't presented or drawn when the PDF is viewed then I'm all ears.


benmccann:

PR #18, opened 2015-10-30
Changed to the getHeight function for fonts so it will return a more accurate height

kalaspuffar:
The getHeight in the fonts gave back approximated heights and in some cases only height the first time the function was called. Tried to clean up the functions and return a more accurate height for each glyph.


benmccann:

PR #19, opened 2016-01-28
PDFBOX-3218 Altering AcroForm merge handling

rsby:
PDFBOX-3218
Altering AcroForm merge handling to avoid merging same fields numerous times


rsby:
Closing to include additional commit

PR #22, opened 2016-02-18
Fix setStrokingColor(PDColor color)

jampukka:
Change nonStrokingColorSpaceStack.size() => strokingColorSpaceStack.size()


THausherr:
Thanks for reporting this bug, I will fix it in PDFBOX-2590. Please give feedback that this solves the problem you were having.

Please close your PR, this is a read only mirror.

PR #23, opened 2016-05-26
Removing the condition that the header must not contain a number

FelipeMGouveia:
Some pdf files that are signed with CMS signature can have a number at the
start of the line beacause of the signature extra content. In those cases
the pdf is fine to be read but the shortcut to ignore lines that starts with
a digit invalidate it.


THausherr:
Please close your request and open an issue in [JIRA](https://issues.apache.org/jira/browse/PDFBOX). This is just a readonly mirror. Please attach a PDF file to explain what you mean.

PR #24, opened 2016-08-17
Needed this exception to be thrown in case invalid fonts are identifi…

uvemulapalli:
…ed while checking the validity of a PDF.


jahewson:
We don't take pull requests, this repo is just a mirror. Our actual source repo is SVN and is hosted by Apache. You're welcome to contribute by opening a JIRA issue and attaching a patch.

Having said that, we won't accept this change. I mean, just look at the comment on the line before:

```
// todo: this too forgiving, but PDFBox has always worked this way for DrawObject
//       some careful refactoring is needed
```

Do you really think that just shoving in `throw e` constitutes "careful refactoring"?

uvemulapalli:
Yeah ?

Kidding :). I do not the reason who/ why that comment was added, but If my PDF contains some invalid fonts, PDF Box supposed to throw an exception back to the user instead of swallowing to its throat like 'LOG.warn(e.getMessage())'.

jahewson:
If PDFBox threw an exception when it encountered an invalid font, about 30% of PDFs would be unusable. Errors are very common in PDF files, and if there's no user action that can be taken to recover from the error then we silently ignore it.

If you're wanting to validate PDF files in some way (and remember, PDFBox silently fixes most errors, so you'll need plenty of your own code), then you should subclass`PDFStreamEngine` and override the `operatorException` method. That's why we put it there :) Our preflight module does exactly that.

PR #25, opened 2016-09-29
PDFBOX-3517 close created RandomAccessFiles

SchwingSK:
simple patch to prevent Windows from keeping the source files locked.


SchwingSK:
Closed because this repository is a read-only mirror.

PR #26, opened 2016-10-03
Update PDFStreamEngine.java

royguo:
There's no need to allocate new ArrayList in `processStreamOperators`. In my test case of a `4.2M` pdf, text extraction reduce from 16 seconds to 14 seconds.


royguo:
typo, will re-create the pull request.

PR #29, opened 2016-10-04
Enhanced ImageType to support BufferedImage.TYPE_3BYTE_BGR

vsadokhin:
The application we are developing converts PDF to an image with TYPE_3BYTE_BGR type. We could do it fine with PDFBox 1.8 but after migration to 2.0 we can't specify this type and have to render an image via PDFBox first, then convert that image to desired TYPE_3BYTE_BGR.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/rendering/ImageType.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/rendering/ImageType.java
@@ -61,6 +61,16 @@ public enum ImageType
         {
             return BufferedImage.TYPE_INT_ARGB;
         }
+    },
+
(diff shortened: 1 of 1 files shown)


THausherr:
This is a read only mirror. Please open a request in JIRA. I believe this was removed here: https://issues.apache.org/jira/browse/PDFBOX-1963

jahewson:
We won't make this change because we have to explicitly add code to support each image type as we often manipulate the image pixels ourselves at a low level. TYPE_3BYTE_BGR is not one of Java2D's  accelerated image formats, so there's no particular reason to want to support this natively. Converting your image afterwards is fine.

Thanks all the same!

PR #31, opened 2016-11-28
Infinite Read Loop

icirellik:
When reading an input file using the BufferedRandomAccesssFile an infinite loop will occur once the buffer is drained and the value `leftover` becomes 0, since the read method will continue to return 0 as the -1 result from `fillBuffer()` is never propagated upstream and any code that is executing in a loop using -1 as an indicator that the file has been fully read will never exit.

An example of where this occurs can be found here:

https://github.com/apache/pdfbox/blob/782ac20dcea7b2e5bb6848f4da9a68c00b4d69a3/fontbox/src/main/java/org/apache/fontbox/ttf/TTFDataStream.java#L264

### DIFF
--- a/fontbox/src/main/java/org/apache/fontbox/ttf/BufferedRandomAccessFile.java
+++ b/fontbox/src/main/java/org/apache/fontbox/ttf/BufferedRandomAccessFile.java
@@ -39,12 +39,12 @@ public class BufferedRandomAccessFile extends RandomAccessFile
     private final byte buffer[];
     private int bufend = 0;
     private int bufpos = 0;
-    
+
     /**
      * The position inside the actual file.
      */
     private long realpos = 0;
-    
+
     /**
      * Buffer size.
      */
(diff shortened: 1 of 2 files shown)


THausherr:
This is a read only mirror. Please close this PR and open an issue in JIRA with your code as a diff. Note that it must be JDK6 compatible, so check whether your java.nio.* methods exist.
https://issues.apache.org/jira/browse/PDFBOX

I'm also interested to read how you hit this problem. Problem with actual PDF, code tool or just looked at something?

icirellik:
I discovered it after upgrading to tika 1.14, one of our large integration tests appeared to be getting stuck. Initial investigation showed that the tika AutoDetector was determining that a couple .dat files should be sent through pdfbox and the thread dumps show that it was stuck in a RUNNABLE state doing font parsing. From there is wasn't to hard to fire up a debugger and see that it was stuck in an infinite loop.

lehmi:
Please close the PR as @THausherr already merged your changes, see [PDFBOX-3605](https://issues.apache.org/jira/browse/PDFBOX-3605) for further details

PR #33, opened 2017-04-09
Support for text positioning 

backslash47:
I am trying to implement text justification for UNICODE encoded texts in project https://github.com/danfickle/openhtmltopdf . Because Tw operator is not supported (is ignored) for unicode strings, I need to calculate and write spacing between characters for every single character. 

I implemented TJ text showing operator, which was not supported by PDPageContentStream.


THausherr:
Please close your PR, this is a read only repository. You can create an issue in [JIRA](https://issues.apache.org/jira/browse/PDFBOX). If you do, please show some test code that would fail (i.e. creates garbage, or something bad or incorrect) without your improvement. I mention this because I haven't yet understood what your PR is about.

PR #34, opened 2017-04-18
throwing errors from operatorException

stevenmcdermott8:


THausherr:
Please close your PR, this is a read only mirror. Feel free to attach a diff to
https://issues.apache.org/jira/browse/PDFBOX-3760
but I won't commit it for the reasons I mentioned there. I do understand your argument, but the majority of users want PDFBox to be lenient.

stevenmcdermott8:
Will do, thank you for the information.

PR #36, opened 2017-04-20
Update PDPageContentStream.java

jknight:
This  function writes bytes as ascii characters directly to the page. 

For use cases where only ascii text needs to be written, this solves a few problems:
1) Avoids the "No glyph for X" issue (eg https://stackoverflow.com/questions/42228567/remove-illegal-characters-from-string-with-pdfbox).
2) While it's possible to pre-process text before using showText(), this means the data will be looped through at least twice since PDFont.encode() loops the text.
3) Preforms better. Everything can be kept as bytes without having to go to text

Benchmarks vs showText():
2000 page pdf: showText(): 579ms / writeAscii(): 419ms
1000 page pdf: showText(): 283ms / writeAscii(): 193ms


THausherr:
This repository is a read only mirror. But the biggest problem is that your code won't work with subsetted type 0 fonts.

jahewson:
> This function writes bytes as ascii characters directly to the page.

Unfortunately it doesn't, as PDF does nit support a literal ASCII encoding. What you'll get is a malformed result that sometimes works for certain fonts, because WinAnsiEncoding != ASCII. Also such encodings only apply to Type 1 fonts, so with OTF and TTF it's a non-starter. You should *not* be writing bytes directly to a content stream.

> 1. Avoids the "No glyph for X" issue (eg https://stackoverflow.com/questions/42228567/remove-illegal-characters-from-string-with-pdfbox).

But this allows the generation of broken PDFs with missing glyphs.

> 2. While it's possible to pre-process text before using showText(), this means the data will be looped through at least twice since PDFont.encode() loops the text.

The looping is unfortunate but it's better than generating PDFs with malformed encodings.

> 3. Preforms better. Everything can be kept as bytes without having to go to text

Performs better... but it's not safe to treat encodings this way. So the two approaches are not equal.

Sorry but this patch just breaks things.

jknight:
Thanks for the feedback. I see this will only work with PDType1 fonts where the text is ascii.
I thought this would accomplish much the same as:

> contentStream.showText(string) 
> --> COSWriter.writeString(font.encode(text), getOutput()); 

where COSWRiter.writeString writes bytes directly to the outputstream, but without needing to have  the text as a string.

jahewson:
> PDType1 fonts where the text is ascii

Yes, sort of, but not really, because PDF doesn't support ASCII, it supports "WinAnsiEncoding", which is Windows-1252. So there's not strictly such thing as "writing an ASCII string to a PDF", in as much as the string must be in some *other*, supported encoding, such as WinAnsiEncoding.

Either way, we still need to check that the font contains the necessary glyphs before we allow the text to be written to the content stream. If we could improve the performance of these checks, that would be a welcome improvement. But bypassing them is not an option.

PR #38, opened 2017-07-01
one-time array sanity check for use with OperatorProcessor

skjolber:
Input for https://github.com/apache/pdfbox/pull/27 (PR not indented for merge). The unit tests seem to work fine with the one-time array. This approach has the limitation that it relies on test coverage. 

Manual inspection seems to verify that the passed array is not saved, however some ArrayList.sublist(..) calls exist; in the arraylist implementation these relay on the parent array. However these are not kept either.


skjolber:
Build is broken in jdk 7, works in jdk 8 - which is okey.

skjolber:
Bump

PR #41, opened 2017-12-08
PDFBOX-4032: Handle correctly write special characters '\n' (LF), '\r…

dudasl:
Pull request for [PDFBOX-4032](https://issues.apache.org/jira/browse/PDFBOX-4032)


marktanner1331:
Looks good to me.

PR #42, opened 2018-01-08
Allow rendering with different X and Y scale

choeflake:
Hi,

We need to render pages sometimes with diffent X and Y scales. But currently there is no public method to do so. Please check the pull request which adds support for separate X and Y scale when rendering to a graphics object.


choeflake:
Issue is PDFBOX-4057

msahyoun:
@choeflake  Could you close the PR as PDFBOX-4057 was implemented?

choeflake:
@msahyoun done!

PR #43, opened 2018-01-25
PDColorSpace ISSUE: Unable to create a valid color space with COSDictionary

santino83:
Unable to create a valid color space when COSBase colorSpace is a COSDictionary with key COSName.COLORSPACE => value a valid color space.

We are run in this issue trying to transform a pdf of a customer of us into jpgs. This pdf has 9 pages, one of them, when ColorSpace is processing, returns a COSDictionary with a single entry, a COSName.COLORSPACE => COSName.DEFAULT_RGB key/value pair. In this situation, PDColorSpace.create throws an exception instead of handle the case. 

We can provide the original pdf that rises the issue, but only in private mode because it's full of personal information and we won't edited it to not risk to lose the original COS configuration


THausherr:
Please send me the file to  tilman at snafu dot de  and mention what confidentiality you need. (keep for myself, share among committers, destroy after fixing bug, or whatever you wish). If possible I'll create a non confidential test case.

PR #44, opened 2018-02-26
When widths not contains the cid, get width from font by code

chunlinyao:
Some PDF use CJK font  without embedded subset displayed incorrectly, the alphabet become wider.

This is result before the patch.
2018-02-26_332x112

This is after patch.
2018-02-26_378x96

The origin pdf.
[dd.pdf](https://github.com/apache/pdfbox/files/1757749/dd.pdf)


THausherr:
This is a read-only mirror, please close this PR and open an issue in [JIRA](https://issues.apache.org/jira/projects/PDFBOX) and attach your .patch file. I tested your patch on my test set, it improves the rendering of several files, but breaks the rendering of the file from [PDFBOX-2966](https://issues.apache.org/jira/browse/PDFBOX-2966), so you may have to fine-tune something.

PR #46, opened 2018-04-13
Enable rendering of Indian languages, by reading and utilizing the GSUB table

paawak:
Implemented proper rendering of Indian languages, which need extensive Glyph substitution. The GSUB table has been read and used effectively to replace some compound words with their respective Glyphs. All tests are passing. I have tested this for the Bengali font. Please review these changes and let me know if it makes sense to incorporate these.

Thanks,
Palash.


THausherr:
Hello, yes, thanks, there has been a need for this for years... however this is a read only repository so you'll have to open an issue in JIRA ( https://issues.apache.org/jira/browse/PDFBOX ). We'll also need an ICLA, please read https://www.apache.org/licenses/icla.pdf , sign and submit it as explained in https://www.apache.org/licenses/#clas .

I only had a quick look (I just woke up), I noticed that there are a few tables and a font. Where do these come from? I see this is only for test so in the worst case we could download it. (Make a text search for "PDFBOX-4038" to see how this is done). We also need the apache header for any new non binary files.

If not in your proposed patch, it would be nice to have an example with a simple text that would not look OK without your patch.

Does your patch change any API in a "breaking" way? If yes it will be used for the trunk only, if not then it could also be used for 2.0.

THausherr:
The font license is OK, we'll only have to include the license text:
https://www.apache.org/legal/resolved.html#category-b

THausherr:
I entered "The quick brown fox jumps over the lazy dog" into Google translate, and the result was "দ্রুত গাঢ় শেয়াল অলস কুকুর জুড়ে জাম্প". Then I created a PDF file with this. With your font, I got a line with incomplete glyphs:
[example-new.pdf](https://github.com/apache/pdfbox/files/1910334/example-new.pdf)
Code:
```
                PDFont font7 = PDType0Font.load(document, new File("Lohit-Bengali.ttf"));
                stream.setFont(font7, 12);
                stream.newLine();
                stream.showText("দ্রুত গাঢ় শেয়াল অলস কুকুর জুড়ে জাম্প");
```

paawak:
Hi Tilman,

Thanks a lot for the quick response. Please find my reply below:

1.  ICLA: Sure, I would send you a signed/scanned copy by 16th April, Monday
2. apache header for any new non binary files: Done
3. example with a simple text: Done: /pdfbox-examples/src/main/java/org/apache/pdfbox/examples/pdmodel/BengaliPdfGenerationHelloWorld.java.
4. Please find attached the before/after pdf files
[Bengali-text-after.pdf](https://github.com/apache/pdfbox/files/1910421/Bengali-text-after.pdf)
[Bengali-text-before.pdf](https://github.com/apache/pdfbox/files/1910422/Bengali-text-before.pdf)
5. Does your patch change any API in a "breaking" way: I don't think so, the only "major" change I did is made PDFont::encode(String text) non-final, and overrode it in PDType0Font
6. Incomplete Bengali text with my changes: Not sure, I just ran it, able to get "almost" complete file, using jdk8.
[a.pdf](https://github.com/apache/pdfbox/files/1910451/a.pdf)

Thanks,
Palash.

paawak:
Created the below Jira to track this change:
https://issues.apache.org/jira/browse/PDFBOX-4189

Thanks,
Palash.

paawak:
Closing the pull request, as the files have been checked in to the trunk by Tilman Hausherr: https://svn.apache.org/r1829710

PR #47, opened 2018-04-22
Refining GSUB system to support new Languages, feature complete Bengali

paawak:
This change set contains some refinement to the code so that GSUB can be supported for any languages with ease. As mentioned in the javadoc for Language enum:
In order to support a new language, you need to add it here and then implement the GsubWorker for the given language and return the same from the GsubWorkerFactory.getGsubWorker(org.apache.fontbox.ttf.CmapLookup, GsubData).

This change set also contains final changes to make Bengali feature complete.

Thanks,
Palash.


paawak:
Closing this PR as this patch has been committed to the trunk by Tilman Hausherr:
https://svn.apache.org/r1830016

PR #48, opened 2018-04-25
Create the GsubWorkerFactory once for a given font, not for every word

paawak:
Create the GsubWorkerFactory once for a given font, not for every word


paawak:
This has to be changed, the implementation is not correct.

PR #50, opened 2018-04-27
minor fix: get gsubWoker from map with font as the key

paawak:
minor fix: get gsubWoker from map with font as the key


paawak:
Commit 1830399 from Tilman Hausherr in branch 'pdfbox/trunk'
[ https://svn.apache.org/r1830399 ]

PR #53, opened 2018-05-03
ABCDEE+Sylfaen are not implemented in PDFBox

bekab95:
May 02, 2018 1:34:54 PM org.apache.pdfbox.pdmodel.font.PDCIDFontType2 <init>
INFO: OpenType Layout tables used in font ABCDEE+Sylfaen are not implemented in PDFBox and will be ignored


THausherr:
This is a read only mirror. Please submit patches to JIRA and close your PR.

You PR changes 628 files - what do you expect????

I suspect you used 2.0 source code on the trunk. Don't do that.

"OpenType Layout tables used in font ABCDEE+Sylfaen are not implemented in PDFBox and will be ignored" is harmless. It is an INFO. Not a warning, not an error. It just says that these tables aren't supported. They may be in the future, a user is working on it.

bekab95:
If I can sllve the issue tell me where is the fonts implemented in the project and I will try also

THausherr:
https://issues.apache.org/jira/browse/PDFBOX-4189

PR #60, opened 2018-11-25
made enum fromString() less error-prone

TorstenF76:
This is no major issue, but when I saw it, i could not leave it as it was. So many hard coded duplicate strings. Error-prone on new enum values.


THausherr:
Hi, this is a read only mirror, you should create an issue in JIRA. However… Although I agree with you that these duplicate strings are a risk, I don't like that your approach uses a loop. It may be slower, and also other people who see this code may come up with an optimization that maybe brings us back to the old code. So unless you can come up with a quote by some java guru who recommends this approach, I think an alternative solution would be to create a test that uses this loop strategy to verify that the strings are equal.

TorstenF76:
Oh man - start think for yourself. Or wait for a java guru. "It may be slower" - there are lots of quotes from java gurus out there that warn you from "premature optimization" - And other tons of quotes that state "maintainable code" is the first duty of every programmer. 

So here's my link to the quotes: https://www.google.com/search?q=premature+optimization 
I wait for your measurement how slow my code is if you - let's say - call this method 1 million times per second - oh, wait - how often is it called? 

Sorry, I won't follow up this contribution - it's your code and it was just an offer to make it better. Please do follow my link.

THausherr:
I committed your patch, thank you. You can close the PR.

brueno:
sorry guys, but why do you not use the static java.lang.Enum.valueOf()-method ?

THausherr:
Heh heh, I tested it and the tests failed. That one works only with the capital names.

PR #67, opened 2018-12-11
2.0

xiexiangyi:
a OutputStream can be  signature then return the OutputStream ?


THausherr:
Please close your PR. This is a read-only mirror. Your PR contains 1901 commits that change 690 files. Please open an issue in JIRA if you have a patch against the trunk, or discuss your argument on the users mailing list first.

It is also unclear whether your text "a OutputStream can be signature then return the OutputStream ?" is a request, a question, a description of a code change, or a comment after a code review.

PR #71, opened 2019-06-26
Change log level for 'Expected number, actual=COSFloat' message

jocko:
It is a bit misleading to log this as an error since this condition doesn't halt or prevent the parser from continuing.


THausherr:
I disagree, this is an error, something is likely lost. If the parser couldn't recover then it would throw an exception.

This is a read only mirror, please close your PR.

PR #72, opened 2019-08-13
PDFBOX-4080 Optimize memory consumption PDLinkAppearanceHandler

gbatalski:
A very naive try to solve the [PDFBOX-4080](https://issues.apache.org/jira/browse/PDFBOX-4080)  problem. Works at least for PDAnnotationLink.


gbatalski:
I would say,  it's done

PR #77, opened 2019-12-03
COSName should be written with same charset as it was read. PDFBOX-4728

oikku:
If read is done using Windows-1252 and write using UTF-8 then PDF containing Windows-1252 encoded XObject dictionry names will be broken after doin load and save

    PDDocument document = PDDocument.load(sourcePath);
    document.save(targetPath); // This will write broken PDF file

### DIFF
--- a/debugger/src/main/java/org/apache/pdfbox/debugger/PDFDebugger.java
+++ b/debugger/src/main/java/org/apache/pdfbox/debugger/PDFDebugger.java
@@ -604,7 +604,6 @@ public class PDFDebugger extends JFrame implements Callable<Integer>, HyperlinkL
         fileMenu.add(printMenuItem);
 
         printDpiMenu = PrintDpiMenu.getInstance().getMenu();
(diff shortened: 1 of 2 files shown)


oikku:
This should also go to 2.0 branch because this bug affects 2.0.17 version.

PascalSchumacher:
As far as I know the pdfbox maintainers do not monitor github pull request.

I would advice you to add a comment with a link to this pull request to the issue (or send an email to the pdfbox developer mailing list).

msahyoun:
@oikku  the issue you are mentioning has been closed and resolved in pdfbox 2.0.9. Please open a new issue in  https://issues.apache.org/jira/projects/PDFBOX describing your issue. Feel free to mention the pull request. It is also advisable to attach a sample PDF and a unit test to reproduce the issue.

Sorry for any inconvenience this may cause but this is only a mirror so the pdfbox maintainers are not notified of this issue

PR #78, opened 2019-12-20
[PDFBOX-3812] Support auto size font for multiline PDTextField

dannymcpherson:
https://issues.apache.org/jira/browse/PDFBOX-3812

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/AppearanceGeneratorHelper.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/AppearanceGeneratorHelper.java
@@ -80,7 +80,12 @@ class AppearanceGeneratorHelper
      * The default font size used for multiline text
      */
     private static final float DEFAULT_FONT_SIZE = 12;    
-    
+
+    /**
+     * The minimum font size used for multiline text auto sizing
+     */
+    private static final float MINIMUM_FONT_SIZE = 4;    
+
     /**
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
As far as I know the pdfbox maintainers do not monitor github pull request.

I would advice you to add a comment with a link to this pull request to the issue (or send an email to the pdfbox developer mailing list).

dannymcpherson:
Thanks for the tip, @PascalSchumacher.  I just emailed their dev mailing list.

msahyoun:
@dannymcpherson  thank you very much for the contribution. @PascalSchumacher  is correct that we don't monitor on a regular basis as this is only a mirror and the issues are being tracked using JIRA. Nevertheless I'll be hopefully able to add the changes later today.

There are some possible enhancements I see
- text can also scale up in order to fit the content area
- a font increment by 1 might be a little large - maybe there could be a setter for the increment step

dannymcpherson:
Thank you, @msahyoun.  I'm working on some changes to support autosizing upward.  I also think I may have found a bug in the logic where it clips at large font sizes.  Stay tuned.

dannymcpherson:
@msahyoun - I just pushed a commit that improves the logic so that we use the largest font size that fits, where the font size is between 4 and 300 (the apparent max used by acrobat pro).

msahyoun:
@dannymcpherson the feature has been fixed in PDFBOX-3812 and release for 2.0.21 and 3.0.0 - can you close the request?

PR #81, opened 2020-02-01
Added support for Tamil language

ksvraja:
Added support for Tamil language  As similar to Bengali language fix - [PDFBOX-4189](https://issues.apache.org/jira/browse/PDFBOX-4189)


PascalSchumacher:
As far as I know the PDFBox team does not monitor pull request. So I suggest you comment on the Jira issue or write to the mailing list to get their attention,

ksvraja:
We will be working on multiple indian languages.  Post that I will put a single pull reqeuest

PR #82, opened 2020-04-01
CCITTFaxDecoderStream Update

Schmidor:
I want to propose the following update of the CCITT decoder.

We had a few fixes since the version from 2016. Futher I later helped to use the decoder in ICEpdf and did some changes for the EncodedByteAlign-Flag.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/filter/CCITTFaxDecoderStream.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/filter/CCITTFaxDecoderStream.java
@@ -4,28 +4,29 @@
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- *     * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Hi Schmidor,

unless I am mistaken the pdfbox committers do not monitor pull requests. 

Better post on the mailing list or to create a Jira issue to ensure that they are aware of your proposal.

THausherr:
Thanks, I've now found this PR and attached it to
https://issues.apache.org/jira/browse/PDFBOX-5057

THausherr:
Thank you for taking the time to merge your changes into our code!

THausherr:
I made some additional changes based on the Sonar report
- check the length of the read bytes. I'm wondering whether it might happen that a valid image has less than 20 bytes, but for now I'll just assume so
- "& 0xff" when using streamData[1] to avoid "int promotion"

THausherr:
> I'm wondering whether it might happen that a valid image has less than 20 bytes, but for now I'll just assume so

It turns out, I hit one that is 19 bytes long.

PR #84, opened 2020-05-22
Add example for custom metadata

rafaelrenanpacheco:


THausherr:
Hi, just as I wanted to add this, I realized that the example file you used is about taking stuff from /Info and adding it to the XML metadata. So you should either alter another example or create a new one.

rafaelrenanpacheco:
Hello @THausherr!

Makes sense, thanks for the feedback.

Best regards,
Rafael Pacheco.

PR #88, opened 2020-10-20
Add a new assertion for isFill

Deee92:
Hello,

I noticed with mutation testing that the behavior of [`org.apache.pdfbox.pdmodel.graphics.state.RenderingMode.isFill()`](https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/state/RenderingMode.java#L91) is not specified.\
If the body of `isFill()` is replaced with `return true;` or `return false;`, no test fails.

Here is a test case that specifies it with a new assertion.

```
@Test
public void testIsFill()
{
    // Arrange
    final RenderingMode objectUnderTest = RenderingMode.FILL;

    // Act
    final boolean retval = objectUnderTest.isFill();

    // Assert result
    Assert.assertEquals(true, retval);
}

```

What do you think?

Thanks!


THausherr:
You're right, and you're welcome. We're always looking to improve our test coverage, thanks.

PR #89, opened 2020-10-26
PDFBOX-5002: fix word detection in PDFTextStripper

SchwingSK:
The problem lied with the fact that maxHeightForLine is kept, even when the text font changes (which is intentional so as not to trigger a new line when there is sub/superscript). This leads in this case to PDFTextStripper merging two lines that should be separate.
The patch assumes that when the current character is separated from the previous one, the maxHeightForLine has to be reset.
This breaks only one test: eu-001.pdf, and it should as the new code correctly detects two lines where there was only one detected before.

(the patch has been tested with mvn clean test on the 2.0.21 branch with commit bdf2ae77e693cc73d4cdeb9a95c6ac2845d11ead applied, as the current 2.0 branch does not pass tests)

The five-spaces rule is purely arbitrary, but gives good results in my tests. It can be changed but should at least be two I feel.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/text/PDFTextStripper.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/text/PDFTextStripper.java
@@ -686,6 +686,17 @@ public class PDFTextStripper extends LegacyPDFStreamEngine
                     {
                         line.add(LineItem.getWordSeparator());
                     }
+                    // if there are too many spaces between last character and current, reset the max line height as the
+                    // font size may have completely changed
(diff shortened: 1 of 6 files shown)


SchwingSK:
After testing with 14646 PDFs, I reduced the five-space rule down to only one (my feeling was wrong ;) ), as it gives even better results, and does not break more TestTextStripper tests.
5 spaces: 965 pages with at least one space fixed out of 14841 pages
1 space: 1083 pages with at least one space fixed out of 14841 pages

PR #93, opened 2020-12-02
Updated comment when rendered image is too large

IntraCherche:
Size image involves w * h * scale squared


THausherr:
why squared?

IntraCherche:
Because the code checks `if ((long) widthPx * (long) heightPx > Integer.MAX_VALUE)` and `widthPx` and `heightPx` are defined as

`int widthPx = (int) Math.max(Math.floor(widthPt * scale), 1);`
` int heightPx = (int) Math.max(Math.floor(heightPt * scale), 1);`

But then the error message uses `widthPt` and `heightPt` (not `widthPx` and `heightPx` as in the if check). So if you rewrite the the if check (just to understand) with `widthPt` and `heightPt` instead of `widthPx` and `heightPx`, you get :
`widthPx * heightPx = widthPt * scale * heightPt * scale` which in turn yields to `widthPt * heightPt * scale * scale ` or `widthPt * heightPt * scale ^ 2` which is the proposed update.

THausherr (approved):

THausherr:
Of course 😂
Good observation!

PR #102, opened 2020-12-28
Add build using GitHub actions

koppor:
This PR adds a [GitHub workflow](https://github.com/features/actions) to enable checks from all JDKs from version 8 to 15.

The output of GitHub actions is better integrated in GitHub. Furthermore, the errors are highlighted better:

grafik

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,54 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


THausherr:
Does this mean that each commit will trigger 8 builds, besides the Jenkins builds and the github travis builds? And same for everybody who has forked the project?

msahyoun:
See https://infra.apache.org/github-actions-secrets.html

THausherr:
"The ASF maxes out its runner allocation quite often, so a project needs to plan carefully to make best use of them for everyone's sake."

So to me this looks as if "no, won't do" is the answer for now.

lehmi:
IMHO it doesn't make that much sense as we don't use github as primary repository, but as readonly mirror. So that those github actions are just another tool to verify our builds **after** a checkin. As Tilmans pointer shows the number of ASF runners are limited so that we shouldn't allocate them at all for a redundant process.

koppor:
OK, got it. I just wanted to make an offer to replace Travis CI. I see ASF having the same limitations as Eclipse Foundation, which I just forgot. Sorry for the noise.

PR #103, opened 2020-12-31
Print stack trace to see why build on github / travis fail here

THausherr:
This is just to test on github. I'll create a JIRA issue if needed.


THausherr:
It's the missing ZapfDingbats font. I'll see whether it's possible to use AFM instead. If not, then I'll abort with a more useful error message.

PR #105, opened 2021-02-28
PDFBOX-4147 Last PDF line should contain '%%EOF'

habbisify:
Fix the problem described in the issue and add a unit test for it.


THausherr:
Why the `pdfStripper`?
Why the change in `PDFontTest.java` (this may be related to a bug that has been fixed)?
Why in `testPDFBox4147` catch the exception and rethrow it?

mkl-public:
As already explained in PDFBOX-4147 itself, the ISO committee responsible showed in the PDF/A specifications that they interpret the PDF specification to allow an `%%EOF` both with and without a following EOL.
Furthermore, there are situations in which an EOL is an advantage, in particular in context of incremental updates.
As there is no issue, therefore, there is nothing to fix.

mkl-public:
Considering https://github.com/apache/pdfbox/pull/105/commits/e37a62198698f73c400f265ae9b4a480a851f92a - is this some kind of homework?

@THausherr 
>Why the `pdfStripper`?

That becomes clear in the intermediary checkin https://github.com/apache/pdfbox/pull/105/commits/ad53c2a1f92f11c85d6cba1076bde49c6bf4f8c7 - a misunderstanding of the issue.

habbisify:
This PR is related to university course me and the other 3 contributors are taking. Part of it was to create a PR for an open issue in any Apache project which fulfills some criteria - however, acceptance of the PR is not mandatory, fortunately. It seems we went for a questionable issue - should it be closed in Jira? As for now it is still marked as a major, open bug.

mkl-public:
I'm merely an (active) PDFBox user, not on the PDFBox team itself, so at best I can advice but neither accept nor reject. So take my words with a grain of salt.
But indeed, in my eyes the issue PDFBOX-4147 should be closed as "not a bug": The PDF/A specifications referencing the PDF specification here clarify how the latter has to be interpreted. And the POSIX reference also points in the same direction.

THausherr:
@habbisify I have closed the JIRA issue. But your PR was created in good faith and I did review it before looking at the JIRA issue discussion and I would have committed it (with slight changes related to my remark earlier today), so thank you nevertheless.
Please send my greetings to your professor. I like this "homework" idea, you get in touch with open source and go "where it hurts" (in the source code) 😂
I'll also review the other PRs later or tomorrow, they seem good.

Please close this PR.

habbisify:
@THausherr Thank you for the effort. I will forward the greetings to the professor!

PR #108, opened 2021-03-17
PDFBOX-xxxx: Add COSName for DocMDP

wdick:
Add COSNames needed to read DocMDP settings
see PDF 32000-1:2008, Table 253 – Entries in a signature reference dictionary


THausherr:
Thanks!

PR #110, opened 2021-03-22
Fix the retrieval of number of components for a JPEG image

matejriseksc:
According to https://docs.oracle.com/javase/8/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html the number of image components should be read from the `markerSequence/sof/@numFrameComponents` if you want to retrieve the number of the whole image.

At the moment the number is being read from the first `Start Of Scan marker segment` (sos) xPath encounters which in our case lead to the wrong classification of an image as a Grayscale instead of RGB resulting in a blank image added to the final PDF document.


matejriseksc:
https://issues.apache.org/jira/browse/PDFBOX-5137

THausherr:
Can you share this image? The oracle text is hard to understand. But I looked into the source code of twelvemonkeys, this confirms what you wrote.

PR #111, opened 2021-03-26
Faster PDImageXObject.applyMask

Schmidor:
Less data copy operations over all. Linewise bulk copy instead of per
pixel.

The softMask / matte calculation might need to be checked. I have only found one PDF in https://issues.apache.org/jira/browse/PDFBOX-4267

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java
@@ -549,7 +549,8 @@ public final class PDImageXObject extends PDXObject implements PDImage
         {
             mask = scaleImage(mask, width, height);
         }
-        else if (mask.getWidth() > width || mask.getHeight() > height)
+
+        if (mask.getWidth() > width || mask.getHeight() > height)
         {
             width = mask.getWidth();
             height = mask.getHeight();
@@ -557,6 +558,7 @@ public final class PDImageXObject extends PDXObject implements PDImage
         }
         else if (image.getType() != BufferedImage.TYPE_INT_ARGB)
         {
+            // always convert to ARGB to allow bulk read / write
             // PDFBOX-4470 bitonal image has only one element => copy into RGB
             image = scaleImage(image, width, height);
         }
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java
@@ -31,7 +31,9 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.ref.SoftReference;
 import java.util.List;
+
 import javax.imageio.ImageIO;
+
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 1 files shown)


THausherr:
Thanks, this looks promising. However the rendering of this file is different
[PDFBOX-2060.pdf](https://github.com/apache/pdfbox/files/6215064/PDFBOX-2060.pdf)

old rendering:
PDFBOX-2060 pdf-1

new rendering:
PDFBOX-2060 pdf-1

difference:
PDFBOX-2060 pdf-1 png-diff

Schmidor:
I missed that not in every case the input is transformed to ARGB, fixed that.

THausherr:
Thanks; there one difference in rendering here (page 14):
[PDFBOX-1359.pdf](https://github.com/apache/pdfbox/files/6217477/PDFBOX-1359.pdf)

Original:
PDFBOX-1359 pdf-14

New:
PDFBOX-1359 pdf-14

Difference:
PDFBOX-1359 pdf-14 png-diff


The difference can't been seen with a human eye so they're minor and I often have such differences myself when making changes. However here I don't understand what happened, there should only be a difference of 1 due to rounding. The differences algorithm (TestPDFToImages.diffImages()) only complains when there is a difference of 3.

Schmidor:
I have added this PDF and page 14 to the TestRendering.renderAndCompare(String fileName).
But that is showing always a difference for me, with both versions from you and also if I switch to a commit before my changes.

THausherr:
If you compared the pre-commits difference against my PNG file then the diff result doesn't mean much, because different jdks  / OS produce different renderings.

I'll try to find out more about the details on the difference, to be sure that there isn't a bug in the diff code.

Schmidor:
Yes, I compared my renderings to your PNG files. Didn't know that :) 
The lines are definately a bit moved between before and after, but I don't see why.

THausherr:
OK I rendered two files at 1500 dpi and yes, there's a difference, so the diff code isn't the cause. The lines in the old codes are smoother. I'll test a few more things, maybe it's related to scaleImage() about interpolation being done or not (despite that the dimensions are identical).

THausherr:
No further ideas from me today; the effect can be seen in PDFDebugger at 1000%, or in the image ( `Root/Pages/Kids/[0]/Resources/XObject/Im1/Resources/XObject/Im0` ) at 100%. I tried to change scaleImage that it copies the pixels directly with getRGB / setRGB and still got the effect.

THausherr:
I tried to copy "image" into an ARGB image with the *old* code and this did not produce the effect, so this would mean scaleImage isn't to blame.

THausherr:
I ran ExtractImages and displayed with IrfanView; on the left, the "old" image, on the right, the "new" image. The green color is the transparent background.
grafik

This rules out that "something" is triggered that happens at a later time.

Schmidor:
Yeah, something when combining image and mask with the formula from PDF.js.
result

I still don't know what I did wrong when changing the old formula from per pixel to per row, but it didn't work and made me look into PDF.js. Seems to be good now.
result

THausherr:
Yeah I think this goes in the same direction as something I suspected this morning, but didn't have the time to test, that the code didn't multiply the matte values (which are 0..1). It works fine now! I'm gonna commit soon.

And it also shows the visual benefit of this "matte" thing.

PR #107, opened 2021-03-01
potential memory leaks and small performance improvements

valerybokov:

PR #112, opened 2021-03-31
split CFFCharset to two classes

valerybokov:
Hi! I found TODO in CFFCharsetCID class and decided to implement it. Hope you find this helpful.
Several notes:
- I created CFFCharsetCID and CFFCharsetType1 (new classes) as package private because CFFCharset is abstract.
Also, I set them as non-abstract because Format0Charset is not abstract. But if you think they should be abstract, I will think about how to change it.
- I extracted the EmbeddedCharset from CFFParser because it is big class. I don't like to use big inner classes and CFFParser has too many inner classes. But if you want you able to move it (make as inner).

### DIFF
--- a/fontbox/src/main/java/org/apache/fontbox/cff/CFFCharset.java
+++ b/fontbox/src/main/java/org/apache/fontbox/cff/CFFCharset.java
@@ -16,46 +16,19 @@
  */
 package org.apache.fontbox.cff;
 
-import java.util.HashMap;
-import java.util.Map;
-
 /**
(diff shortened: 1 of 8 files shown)


THausherr:
Andreas is currently busy with refactoring CFF, so it's better to comment there, so that you don't interfere with each other (this applies also to your recent comment in #107)
https://issues.apache.org/jira/browse/PDFBOX-5143

lehmi:
I'm going to look into this as well as the others proposed changes in the CFFParser area

PR #114, opened 2021-04-10
PDFBOX-4073 Choosable Coordinate-Unitsystem

habbisify:
This pull request is discussed in Jira ticket:
https://issues.apache.org/jira/browse/PDFBOX-4073

Our take on this: There could be a need to work with millimetres or inches instead of points. @THausherr commented that he uses PDFBox with points and adjusts the parameters via trial and error until he likes what he sees. However, our take on this topic is that there are documents which are strictly standardized to have certain measures, such as offset from the paper edges, in millimetres or inches.

Another concern by @lehmi was that the original idea of the reporter, "a central setting of an unitconverter so that every coordinate is converted deep inside shortly before rendering", is problematic since it will ultimately lead to more complicated and error prone code. We agree with this point.

Therefore, instead of implementing an internal unitconverter setting, we simply implemented a convertUnit method inside PDAbstractContentStream to do the unitconversion from mm or inches to points. This will not complicate other code or make it more error prone - it is just a stand-alone method to do the conversion. Basic test was also implemented to check that the unitconversion works as expected with the convertUnit method.

Please note, this PR is a mandatory part of the university course we are taking. However, acceptance of the PR is not required. This is also the last submission of our group during this course.

What do you think?


THausherr:
I can't remember that said that but I see I wrote it in the issue. And yes it's true 😂. I just had to create a PDF a week ago based on an existing PDF from a software that will be replaced. I used PDFDebugger and moved the mouse and saw the coordinates.

The current code is just a tool that converts a list of numbers, that might make it more verbose, i.e. add to a list, convert, retrieve from the list.

PR #117, opened 2021-05-12
PDFBOX-5191 | isEmbeddingPermitted() is too restrictive on TTFs with OS2 table version

larrylynn-wf:
### PROBLEM
New versions of pdfbox are too restrictive and refuse to embed some fonts with OS2 tables version 0-2 that have multiple permission but set.

### SOLUTION
tweak the isEmbeddingPermitted() method to return the proper boolean when multiple permission bits are set (like 0110 = 6).  Add unit tests to exercise all legal permission combinations

### DIFF
--- a/pdfbox/pom.xml
+++ b/pdfbox/pom.xml
@@ -75,6 +75,12 @@
             <artifactId>jbig2-imageio</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.3.3</version>
+            <scope>test</scope>
+        </dependency>
         <!-- For legal reasons (incompatible license), these three dependencies below
         are to be used only in the tests and may not be distributed. 
         See also LEGAL-195 -->
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/TrueTypeEmbedder.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/TrueTypeEmbedder.java
@@ -142,14 +142,7 @@ abstract class TrueTypeEmbedder implements Subsetter
         {
             int fsType = ttf.getOS2Windows().getFsType();
             int maskedFsType = fsType & 0x000F;
-            // use two's complement arithmetic to check if fsType is a power of 2
-            // all legal fsType combinations for fonts having a version 3 OS2 table are powers of 2
-            // there are more legal combinations for versions 0-2, but the most permissive pit takes precedence
-            boolean powerOfTwo = maskedFsType != 0 && (maskedFsType & (maskedFsType - 1)) == 0;
-            if (
-                    powerOfTwo &&
-                    (fsType & OS2WindowsMetricsTable.FSTYPE_RESTRICTED) == OS2WindowsMetricsTable.FSTYPE_RESTRICTED
-            )
+            if (maskedFsType == OS2WindowsMetricsTable.FSTYPE_RESTRICTED)
             {
                 // restricted License embedding
                 return false;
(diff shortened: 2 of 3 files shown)


pdfbox/pom.xml
```diff
@@ -75,6 +75,12 @@
             <artifactId>jbig2-imageio</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
```
larrylynn-wf, line 80: I didn't see any existing tests using a mocking library.  I'm not sure if this is intentional.  I'm also not sure if pdfbox maintainers have a policy against adding new dependencies.  If mockito is undesirable for some reason, I'll have to come up with a different way to test my proposed code change.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/TrueTypeEmbedder.java
```diff
@@ -136,13 +136,20 @@ public final void buildFontFile2(InputStream ttfStream) throws IOException
...
      */
-    private boolean isEmbeddingPermitted(TrueTypeFont ttf) throws IOException
+    protected boolean isEmbeddingPermitted(TrueTypeFont ttf) throws IOException
     {
         if (ttf.getOS2Windows() != null)
         {
             int fsType = ttf.getOS2Windows().getFsType();
-            if ((fsType & OS2WindowsMetricsTable.FSTYPE_RESTRICTED) ==
-                             OS2WindowsMetricsTable.FSTYPE_RESTRICTED)
+            int maskedFsType = fsType & 0x000F;
```
larrylynn-wf, line 144: For this PR, I branched off of the trunk.  I believe this is essentially the version 3.0.x pre-release codebase.  However, this method is identical in trunk and the current stable release version, 2.0.23.  So, I think this patch will work for both the 2.0.x codebase and the 3.0.x codebase.

pdfbox/src/test/java/org/apache/pdfbox/pdmodel/font/TestFontEmbedding.java
```diff
@@ -371,4 +376,116 @@ void testReuseEmbeddedSubsettedFont() throws IOException
...
+        }
+    }
+
+    /**
+     * Test that we validate embedding permissions properly for all legal permissions combinations
+     *
+     * @throws IOException
+     */
+    @Test
+    void testIsEmbeddingPermittedMultipleVersons() throws IOException
```
larrylynn-wf, line 402: This test is compatible with the version 3.0.x codebase.  However, it looks like the testing frameworks have changed significantly between version 2.0.x and 3.0.x.  This test won't work as-is in the 2.0.x codebase.  

I think I need to have a discussion with the pdfbox maintainers about how they want to handle test coverage for my proposed change.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/TrueTypeEmbedder.java
```diff
@@ -136,13 +136,20 @@ public final void buildFontFile2(InputStream ttfStream) throws IOException
...
+            // use two's complement arithmetic to check if fsType is a power of 2
+            // all legal fsType combinations for fonts having a version 3 OS2 table are powers of 2
+            // there are more legal combinations for versions 0-2, but the most permissive pit takes precedence
+            boolean powerOfTwo = maskedFsType != 0 && (maskedFsType & (maskedFsType - 1)) == 0;
```
larrylynn-wf: I think that it's worth pointing out that this code will fail open on an illegal configuration.

If a font with a version 3 OS2 table has multiple permission bits set, this is an illegal configuration because version 3 explicitly specifies that permission bits should be mutually exclusive.  Suppose we are processing a font with version 3 and permission set to 0110.  This code would allow embedding of the font with the illegal permission configuration.  Essentially, it will fail open.

I think this is OK because in this case, the creator of the font has not made their wishes unambiguous and clear.  What if the font creator wanted us to respect the other permission bit and not FSTYPE_RESTRICTED.  I think failing open is OK, and it essentially puts the onus on font creators.  If they want to lock their fonts down and prevent embedding, it is up to them to provide a legal fsType configuration that makes it clear that that is their intention.

I can see other points of view on this issue though, so I thought it was at least worth pointing out for possible discussion.

PR #122, opened 2021-06-18
Update SampledImageReader.java

gunnar-ifp:
This fixes an issue with a missed logged error + loop breaking if an erroneous read falls onto a skipped row.

It also improves the speed at least by factor twice  (due to sign test instead of bit masking and boolean flag for subsampling).
It will be even faster with regions, as there was a poor design choice of full inner loop computations even if x was below startX. 

(The method wasn't very slow to begin with, though.)

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/SampledImageReader.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/SampledImageReader.java
@@ -171,7 +171,7 @@ final class SampledImageReader
     }
 
     /**
-     * Returns the content of the given image as an AWT buffered image with an RGB color space.
+     * Returns the content of the given image as a newly created AWT buffered image with an RGB color space.
      * If a color key mask is provided then an ARGB image is returned instead.
      * This method never returns null.
      * @param pdImage the image to read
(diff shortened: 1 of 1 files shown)


THausherr:
Do you have a file that shows the speed increase / how did you test? I tried with the file from https://issues.apache.org/jira/browse/PDFBOX-2123 and didn't really notice a difference with PDFDebugger (but it's difficult to make benchmarks).

gunnar-ifp:
Yeah as I wrote in the initial comment, it's not really slow. I simply timed the method itself in the sourcecode with System nanoseconds and for the big mask it went down from 55 to something 25ish. Once you go through a few pages of that Alfa catalog PDF you can see consisten times. As I said it wasn't really slow. It was just very wasteful in the original version with not calculating start offsets and all.

And the error if readLen!=rowLen was not caught with subsampling if the line was left out.

THausherr:
This patch is against the 2.0 branch. Can you modify it so that it is against the trunk? (It does work with 2.0 and there are no regressions, but it is easier for me to test with the trunk).

THausherr:
Maybe it isn't needed, it's probably just a block and I can do this myself easily.

THausherr:
Fixed in https://issues.apache.org/jira/browse/PDFBOX-5224, thank you!

PR #135, opened 2021-11-07
Attempting to fix issues with printing

trumpetinc:
see "Guidance on heap space error when printing document" discussion on user listserv


THausherr:
Why did you change the equals and the hash, is this for optimizing? (Assumption that xform is only scale, never shear)

I tried the scale change from your last commit, this does not work well when setting MAXEDGE at 500. The "Schleuse" file looks very different. The problem is that the pattern gets shorter and is scaled so the repetition is visible, while a crop has no effect as long as the cropped part is outside the page. This can also be seen on page 7 of the file from PDFBOX-3653.pdf .

trumpetinc:
Yes - it is an optimization.  I'm pretty sure you can re-use the cached
value if there is only a translation.  But please sanity check my logic on
this.

This is not the most important optimization for printing - by far the most
important optimization is caching of the PageDrawer.


And switching to the weak *valued* cache in TilingPatternFactory is
important as the current cache is going to evict very rapidly.

- K

Kevin Day

*trumpet**p| *480.961.6003 x1002
*e| ***@***.***
www.trumpetinc.com <http://trumpetinc.com/>
LinkedIn <https://www.linkedin.com/company/trumpet-inc.>| Trumpet Blog
<http://trumpetinc.com/blog/>| Twitter  <https://twitter.com/trumpetinc>

Proud to be Great Place To Work
<https://www.greatplacetowork.com/certified-company/7012667> certified
since 2019

PR #139, opened 2022-01-15
Replace finalize() with Cleaner

rototor:
Finalizers (method finalize()) are going to be deprecated for removal with JDK 18. See https://openjdk.java.net/jeps/421 for details.

The best way to replace the finalize() methods is by using the JDK 9 java.lang.ref.Cleaner. As PDFBox 3 targets JDK 8 this can not be used directly.

The attached patch implements a Cleaner using finalizers for JDK <= 8 and using java.lang.ref.Cleaner by reflection for JDK 9+. 

The two remaining finalize() implementing classes are migrated to the new Cleaner.

I’m not really happy with the name and package org.apache.fontbox.util.PdfBoxInternalCleaner of the cleaner. Maybe you have an idea for a better place and name.

In theory this patch could be back ported to PDFBox 2, but I’m not sure if this is worth the risk.


rototor:
See also the JIRA Ticket https://issues.apache.org/jira/browse/PDFBOX-5362

PR #140, opened 2022-01-16
PDFBOX-5363: Don't log warnings if there are not fonts to cache

famod:
I kept `saveDiskCache()` out of the `if` block so that the cache file reflects reality (in case it was there already with whichever content).


famod:
Superseded by https://github.com/apache/pdfbox/commit/7c3fa21b5597b5a761a9c05f89652437322e1813

PR #147, opened 2022-10-31
[PDFBOX-5526] Adding subsampling for masks

velyan:
See[ Jira ticket ](https://issues.apache.org/jira/browse/PDFBOX-5526?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17618666#comment-17618666)

When `PDImageXObject.getImage()` gets invoked with subsampling and region, internally it loads the entire mask into memory: https://github.com/apache/pdfbox/blob/961c052d52dd9ab2dd3d7cd762a5046e5cc85a91/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java#L465

and applies the entire mask on the subsampled image.

Which is extra work and can cause `OOM` exceptions. 

The proposed optimisation is to pass `region` and `subsampling` params to `PDImageXObject.getOpaqueImage() `here - https://github.com/apache/pdfbox/blob/961c052d52dd9ab2dd3d7cd762a5046e5cc85a91/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java#L548

such that masks are also subsampled before applying them...

### DIFF
--- a/examples/src/main/java/org/apache/pdfbox/examples/signature/CreateVisibleSignature.java
+++ b/examples/src/main/java/org/apache/pdfbox/examples/signature/CreateVisibleSignature.java
@@ -32,7 +32,7 @@ import java.util.Calendar;
 import org.apache.pdfbox.Loader;
 import org.apache.pdfbox.cos.COSName;
 import org.apache.pdfbox.io.IOUtils;
-import org.apache.pdfbox.io.MemoryUsageSetting;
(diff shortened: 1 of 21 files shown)


THausherr (commented):
You're not passing the region despite having it as a parameter

velyan:
> You're not passing the region despite having it as a parameter

Thanks, fixed!

THausherr (approved):

PR #148, opened 2023-01-09
Fix meta markup in HTML generation

guillaume-alvarez:
The current code generates incorrect HTML, a double-quote is never closed in
```
<meta http-equiv="Content-Type" content="text/html; charset="UTF-8">
```
The meta.content attribute does not need internal double-quotes so I removed it.


THausherr:
Thank you!

PR #149, opened 2023-01-10
Allow negative struct parent keys in PDFMergerUtility

bernhardf-ro:
PDFMergerUtility expects structural parent tree keys to be non-negative.
However, negative values don't seem to be forbidden by the specification and are accepted by validators.
This patch adapts the class to handle negative struct parent keys correctly.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
@@ -795,6 +795,7 @@ public class PDFMergerUtility
                         if (!srcNumberTreeAsMap.isEmpty())
                         {
                             mergeStructTree = true;
+                            destParentTreeNextKey -= Collections.min(srcNumberTreeAsMap.keySet()); // allow for negative struct parent values
                         }
                     }
                 }
@@ -1498,8 +1499,8 @@ public class PDFMergerUtility
         List<PDAnnotation> newannots = new ArrayList<>(annots.size());
         annots.forEach(annot ->
         {
-            int structParent = annot.getStructParent();
-            if (structParent >= 0)
+            int structParent = annot.getCOSObject().getInt(COSName.STRUCT_PARENT, Integer.MIN_VALUE); // allow for negative struct parent values
+            if (structParent > Integer.MIN_VALUE) // allow for negative struct parent values
             {
                 annot.setStructParent(structParent + structParentOffset);
             }


THausherr:
Do you have a PDF where this happens?

bernhardf-ro:
Here is a document to verify the issue. (Sorry, I forgot to add that to the issue in the first place.)
[merge.pdf](https://github.com/apache/pdfbox/files/10393126/merge.pdf)
It is a valid PDF/UA-1, according to PDF Accessibility Checker 2021.
Appending it to itself using PDFMergerUtility results in a document that is not valid PDF/UA. ("Structural parent tree" issue)
With the patch applied the merge result is valid PDF/UA-1.

pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
```diff
@@ -1498,8 +1499,8 @@ private void updateStructParentEntries(PDPage page, int structParentOffset) thro
         List<PDAnnotation> newannots = new ArrayList<>(annots.size());
         annots.forEach(annot ->
         {
-            int structParent = annot.getStructParent();
-            if (structParent >= 0)
+            int structParent = annot.getCOSObject().getInt(COSName.STRUCT_PARENT, Integer.MIN_VALUE); // allow for negative struct parent values
```
lehmi, line 1502: This doesn't change anything but the default value if STRUCT_PARENT is null. If Integer.MIN_VALUE makes more sense than -1 the getter should be changed rather than this piece of code

bernhardf-ro: I intentionally kept the patch to one class and as simple and minimal as possible, e.g. not renaming variables. So feel free to adapt it and let me know if you have any questions.

lehmi: So the question is, do we need to change the default value if the value isn't set?

bernhardf-ro: That would be an API change which would break all specific (i.e. positiv or negative equals) checks for -1 in all integrations.
If a one-parameter method that defaults to `Integer.MIN_VALUE` is necessary (IMHO it is not), it would have to be a new one, e.g. `getIntSigned(COSName)` (and similar for array, possibly for other number types).
The only change in this regard that I consider necessary is improving the API documentation of `getInt(COSName)` (and similar methods) to clarify that they cannot be used if the result may be negative.

PR #151, opened 2023-01-18
PDResource font cache improvement

knazarenkov:
Given:
PDF with AcroForm with N text fields and M fonts loaded.

Problem: 
When PDAcroForm.getDefaultResources() is called, new PDResource with new direct font cache is created. So whenever PDTextField.setValue() is called, PDFontFactory.createFont(), which is expensive operation, is called M + 1 times - M times in AppearanceGeneratorHelper and once when parsing appearance string in PDDefaultAppearanceString. So setting default appearance for N fields would cause fonts to be loaded  N * (M + 1) times.

This PR's intention is to introduce AcroForm-level direct font cache that would be reused in each instance created by PDAcroForm.getDefaultResources()

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDResources.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDResources.java
@@ -43,7 +43,7 @@ import org.apache.pdfbox.pdmodel.graphics.PDXObject;
 
 /**
  * A set of resources available at the page/pages/stream level.
- * 
+ *
  * @author Ben Litchfield
  * @author John Hewson
  */
@@ -51,11 +51,10 @@ public final class PDResources implements COSObjectable
 {
     private final COSDictionary resources;
     private final ResourceCache cache;
-    
+
     // PDFBOX-3442 cache fonts that are not indirect objects, as these aren't cached in ResourceCache
     // and this would result in huge memory footprint in text extraction
-    private final Map <COSName,SoftReference<PDFont>> directFontCache = 
-            new HashMap<COSName, SoftReference<PDFont>>();
+    private final Map <COSName,SoftReference<PDFont>> directFontCache;
 
     /**
      * Constructor for embedding.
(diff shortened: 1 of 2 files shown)


pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDResources.java
```diff
@@ -95,6 +96,24 @@ public PDResources(COSDictionary resourceDictionary, ResourceCache resourceCache
...
         cache = resourceCache;
+        directFontCache = new HashMap<COSName,SoftReference<PDFont>>();
+    }
+
+    /**
+     * Constructor for reading.
+     *
+     * @param resourceDictionary The cos dictionary for this resource.
+     * @param resourceCache The document's resource cache, may be null.
+     */
```
THausherr, line 108: please add a good javadoc line here, for the parameter

THausherr:
Yeah I think this makes sense. Builds pass.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDResources.java
knazarenkov: While adding javadoc I realised that mutability of cache cannot be deduced from method signature by reader - someone could try to use immutable map as parameter, so new type is introduced that is mutable by design - PDFontCache.
Javadoc added also

lehmi: I don't like the idea to introduce a new type just to avoid that people might follow the bad idea to pass an immutable map for a cache. IMHO if the do so, they have to deal with an exception.

knazarenkov: Removed introduced type

knazarenkov:
@THausherr @lehmi Any updates on reviewing PR?

lehmi:
I've opened a JIRA ticket, see https://issues.apache.org/jira/browse/PDFBOX-5564

mslowiak:
@lehmi, @THausherr - Is there any ETA when `2.0.28` will be released with this fix?

PR #153, opened 2023-01-22
PDFBOX-4189: Add a method to get GSUB data for specific script tag

Toparvion:
This is a proposal for overcoming current GSUB table handling limitation of FontBox library. It is currently not possible to read GSUB table data for any language other than Bengali.

This PR introduces 2 new methods to `org.apache.fontbox.ttf.GlyphSubstitutionTable` class:

1. `getSupportedScriptTags` to fetch a read-only view of the `scriptList` map keys;
1. `getGsubData(String)` to obtain `GsubData` for given script tag (that can be determined by the method above).

In order to make it possible to read GSUB data without actual glyph substitution, the PR also changes the behavior of `org.apache.fontbox.ttf.gsub.GsubWorkerFactory`:

- **Before** the change it used to throw an exception for any non-supported language (in fact - for any one other than Bengali);
- **After** the change it would return a default `GsubWorker` implementation that doesn't perform actual glyph transformations (substitutions) but emits a WARN message into the log.

The PR introduces primarily additive changes. The previous behavior of obtaining GSUB data for the first detected supported language (and its caching) has been kept intact.


fontbox/src/main/java/org/apache/fontbox/ttf/GlyphSubstitutionTable.java
```diff
@@ -701,6 +703,42 @@ public GsubData getGsubData()
...
+     * a <a href="https://learn.microsoft.com/en-us/typography/opentype/spec/scripttags">script
+     * tag</a> for which the data is needed
+     * @return GSUB data for the given script or {@code null} if no such script in the font
+     */
+    public GsubData getGsubData(String scriptTag)
+    {
+        ScriptTable scriptTable = scriptList.get(scriptTag);
+        if (scriptTable == null)
+        {
+            return null;
```
Toparvion, line 725: It could seem reasonable to align this behavior with the sibling method and return `GsubData.NO_DATA_FOUND` from here, but I find it confusable for the user because without thorough reading of JavaDoc (which is normal) it would be hard to understand that the returned result is actually "no result", while with `null` it is definitely clear.

fontbox/src/main/java/org/apache/fontbox/ttf/gsub/DefaultGsubWorker.java
```diff
@@ -0,0 +1,43 @@
...
+class DefaultGsubWorker implements GsubWorker
+{
+    private static final Log LOG = LogFactory.getLog(DefaultGsubWorker.class);
+
+    @Override
+    public List<Integer> applyTransforms(List<Integer> originalGlyphIds)
+    {
+        LOG.warn(getClass().getSimpleName() + " class does not perform actual GSUB substitutions. "
+                + "Perhaps the selected language is not yet supported by the FontBox library.");
+        return originalGlyphIds;
```
lehmi: We should return either an unmodifiable version or a copy of the origin list, as the caller doesn't expect to get back the provided instance of the list. This may lead to unwanted changes of the origin list.

Toparvion: Indeed. Thank you for pointing to this!

In this particular case I've preferred to make the result unmodifiable as it would clearly inform the user about incorrect modification attempt (unlike copying which would "swallow" the modification thus hiding a probable bug).

Toparvion:
✅ The fix has been [integrated](https://svn.apache.org/viewvc?view=revision&revision=1907078) into Apache SVN repository by @lehmi within issue [PDFBOX-4189](https://issues.apache.org/jira/browse/PDFBOX-4189).

PR #157, opened 2023-03-17
Fix log issues

xzel23:
I fixed some smaller problems with logging in trunk:
- GlyphRenderer conditions and logging calls did not match leading to unnecessary string formatting when log level is debug
- BidiTest logger was initialized with the wrong class
- InvalidFileTest used a non-static, non-final logger without reason


xzel23:
created [PDFBOX-5572](https://issues.apache.org/jira/browse/PDFBOX-5572) for this.

xzel23:
Has been applied in SVN by Tilmann Hausherr.

PR #158, opened 2023-03-17
small code cleanups

xzel23:
- in some places, unnecessary explicit boxing/unboxing was done
- I replaced some XXX.valueOf(String) calls with the corresponding XXX.parseXXX(String) calls. XXX.valueOf() always returns a boxed value, but when that is assigned to a primitive, the boxing/unboxing is needless.
- added generics information or the diamond operator ('<>') so that IDEs keep silent about using raw types
- replaced explicit generic parameters with diamond operator where applicable
- changed String.indexOf(...)==0 with String.contains(...) in one place


xzel23:
I will split this up into smaller chunks

PR #159, opened 2023-03-18
fix unnecessary boxing/unboxing

xzel23:
Most of the changes make the code just more readable and the compiler does auto(un)boxing where necessary.

Replacing `Boolean.valueOf()` and `Float.valueOf()` on the other hand completely avoids a boxing/unboxing cycle because in both cases the result is stored in a primitive.


THausherr:
Please close this one (I can close it through the commit message but forgot, twice)

PR #164, opened 2023-06-29
New tests for `COSName.writePDF`

Deee92:
Hello!

I tried a few command-line tools on some PDF documents, and found that [`org.apache.pdfbox.cos.COSName.writePDF(OutputStream)`](https://github.com/apache/pdfbox/blob/b5658f36d525961a2d086acf7c2c5a188e1a1c5b/pdfbox/src/main/java/org/apache/pdfbox/cos/COSName.java#L774) was invoked during the execution.

Here are two tests that verify the invocation of `java.io.OutputStream.write(int)` within `writePDF`. The `OutputStream` object is mocked within the tests.

The following test verifies the parameters with which `write` should be invoked within `writePDF`:

```
@Test
public void testBytesWrittenToOutputStreamForCOSNameType() throws IOException
{
    // Arrange
    COSName cosNameType = COSName.TYPE;
    OutputStream mockOutputStream = Mockito.mock(OutputStream.class);

    // Act
    cosNameType.writePDF(mockOutputStream);

    // Assert
    verify(mockOutputStream, atLeastOnce()).write(47);
    verify(mockOutputStream, atLeastOnce()).write(84);
    verify(mockOutputStream, atLeastOnce()).write(121);
    verify(mockOutputStream, atLeastOnce()).write(112);
    verify(mockOutputStream, atLeastOnce()).write(101);
}
```

The following test verifies that `write` should be called exactly five times within `writePDF`:

```
@Test
public void testNumberOfBytesWrittenToOutputStreamForCOSNameType() throws IOException
{
    // Arrange
    COSName cosNameType = COSName.TYPE;
    OutputStream mockOutputStream = Mockito.mock(OutputStream.class);

    // Act
    cosNameType.writePDF(mockOutputStream);

    // Assert
    verify(mockOutputStream, times(5)).write(anyInt());
}
```

What do you think?

Thanks!

### DIFF
--- a/pdfbox/src/test/java/org/apache/pdfbox/cos/TestCOSName.java
+++ b/pdfbox/src/test/java/org/apache/pdfbox/cos/TestCOSName.java
@@ -17,23 +17,25 @@
 package org.apache.pdfbox.cos;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.*;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 
 import org.apache.pdfbox.Loader;
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDPage;
 import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
 
 class TestCOSName
 {
     /**
      * PDFBOX-4076: Check that characters outside of US_ASCII are not replaced with "?".
-     * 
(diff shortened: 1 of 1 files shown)


THausherr:
I don't see how this improves test coverage; at most, this shows an interesting example on how to use mockito.

Deee92:
Hi @THausherr, that's a fair point!
However, the integer parameters (`47`, `84`, `112`, `121`, `101`) with which `OutputStream.write(int)` is called within `COSName.writePDF(OutputStream)`, as well as the number of times it should be called (`5`) correspond to actual behavior during execution.
Additionally, I did some mutation analysis, and found that the two tests can kill mutants introduced within `writePDF`.
These make them potentially useful as regression tests. 🚀

lehmi:
Your is a valid test, but it won't increase out test coverage as that method in question is covered by other tests already.

PR #165, opened 2023-09-27
PDFBOX-5687 - PDFBox 3.0 OSGi bundle requires sun.java2d.cmm.kcms package

rombert:
Disable generating Import-Package statements for "Class.forName" calls completely.


rombert:
Closed with a different solution, see Jira issue for details.

PR #167, opened 2023-11-18
extract class refactoring.

Vivek504:
### Extract Class Refactoring

**Location of the class:** org.apache.pdfbox.examples.pdmodel.CreatePatternsPDF

**Reason:** Imperative Abstraction - The tool detected the smell in this class because it looks like that an operation is turned into a class. This class contains only the following public method: main

**Logic of the refactoring:** New class (PatternDrawer) is created to refactor the code and refactored code separates the responsibilities of two classes (CreatePatternsPDF and PatternDrawer).


THausherr:
This is an example class. There is no need to refactor it and create an additional class.

Vivek504:
Thanks for your quick feedback. I'll try to contribute if there is a need of refactoring any other class instead of Example.

PR #172, opened 2023-11-19
push down method refactoring.

Vivek504:
### Push Down Method Refactoring

**Reason:** Rebellious Hierarchy - Method of subclass reject the methods provided by its supertypes because supertype's method only returns the empty object. The method that reject the suptertype's method is: readTable
**Logic of the Refactoring:** Removed the method from the super-type.


THausherr:
Your commit removes a part of the API.

PR #173, opened 2023-11-28
PDFBOX-5715: Lines vanish when printing on MacOS

AlexanderScherbatiy:
This a fix for the [PDFBOX-5715](https://issues.apache.org/jira/browse/PDFBOX-5715) `Lines vanish when printing on MacOS`.

PageDrawer class sets clip the same as drawn path on line https://github.com/apache/pdfbox/blob/6441a8b6117c7ac9963491de79c8c5f07e4a52df/pdfbox/src/main/java/org/apache/pdfbox/rendering/PageDrawer.java#L998

When the graphics scale is less than 1.0 and the path stroke width is less than 1.0 then the clip width or height is less than 1.0 and it can be rounded in the device space to have a zero area.

The issue with the fop project is reproduced on MacOS because CPrinterJob sets graphics scale to 1 for Printable, PDFBox adds scale `0.4842` in code:
https://github.com/apache/pdfbox/blob/2b60e17e1db7bfc29d32ba33530d305c6441aaa1/pdfbox/src/main/java/org/apache/pdfbox/printing/PDFPrintable.java#L213
where imageable area is `[288, 432]`, crop box area: `[0.0,0.0,594.72,792.0]` and the scaleX is `288/594.72 = 0.4842`, and the linePath stroke is `0.283`.

On Linux RasterPrinterJob sets the graphics scale to `4.16` which is enough to properly draw the line path in the device space.
The RasterPrinterJob calculates `x/yScale` as [ getX/YRes() / 72.0](https://github.com/openjdk/jdk/blob/4d50df63b6ea76fd6a6c387593c3d6bc9429007b/src/java.desktop/share/classes/sun/print/RasterPrinterJob.java#L2175) where the default [getX/YRes is 300](https://github.com/openjdk/jdk/blob/4d50df63b6ea76fd6a6c387593c3d6bc9429007b/src/java.desktop/share/classes/sun/print/PSPrinterJob.java#L147).

The fix checks if the graphics scale is less than 1.0 and the scaled linePath bounds width or height is less than 1.0 and then enlarges the clip bounds to cover at least 1 point plus 0.5 on one and another side in the device space to allow to draw the linePath inside the clip.
The linePath can consists from different lines but when its bounds width or height is less than 1.0 it seems safe to use a rectangle as a clip instead of the real path.

For performance reasons the fix first checks if the graphics scale is less than 1.0 and if it is not the case the initial clip is used. It still can be the case that the scale is larger than 1.0 but the line stroke width is less than 1.0 so the total stroke width in the device space is less than 1.0.

The tests pass on Linux with the fix.
There are failing tests on MacOS but as I can see the same tests fail on MacOS even without the fix:

The only test which uses the graphics scale less than 1.0 on Linux is `org.apache.pdfbox.tools.imageio.TestImageIOUtils` but its linePath bounds widht and height are greater than 1.0.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/rendering/PageDrawer.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/rendering/PageDrawer.java
@@ -52,6 +52,7 @@ import java.awt.image.WritableRaster;
 import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -103,6 +104,11 @@ import org.apache.pdfbox.pdmodel.interactive.annotation.PDAnnotationUnknown;
 import org.apache.pdfbox.pdmodel.interactive.annotation.PDAppearanceDictionary;
 import org.apache.pdfbox.util.Matrix;
 import org.apache.pdfbox.util.Vector;
+
+import static java.awt.geom.AffineTransform.TYPE_IDENTITY;
+import static java.awt.geom.AffineTransform.TYPE_TRANSLATION;
+import static java.awt.geom.AffineTransform.TYPE_FLIP;
+import static java.awt.geom.AffineTransform.TYPE_MASK_SCALE;
 
 /**
  * Paints a page in a PDF document to a Graphics context. May be subclassed to provide custom
@@ -995,7 +1001,7 @@ public class PageDrawer extends PDFGraphicsStreamEngine
             if (!linePath.getPathIterator(null).isDone())
             {
                 // PDFBOX-4949 / PDF.js 12306: don't clip if "W n" only
-                getGraphicsState().intersectClippingPath(linePath);
+                getGraphicsState().intersectClippingPath(adjustClip(linePath));
             }
 
             // PDFBOX-3836: lastClip needs to be reset, because after intersection it is still the same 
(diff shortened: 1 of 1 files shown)


THausherr:
Interesting, I tested on windows and it makes no difference at 100% (which your text mentions) but it makes a difference at 50% in PDFDebugger. I'll wait for feedback from Trey. Update: it works for him.

tresf:
>  I'll wait for feedback from @tresf

@THausherr sorry, I should have subscribed to notifications.  Yes, by running the PDFBOX-5715 test case against this PR, it directly fixes the symptom.

* BEFORE: `border.pdf` has random lines missing
* AFTER: `border.pdf` shows all lines

>  I tested on windows and it makes no difference at 100% (which your text mentions) but it makes a difference at 50% in PDFDebugger.

I've only tested on macOS, but since the test provided in PDFBOX-5715 uses a page size of 4x6 (roughly 50% of a US Letter or A4 Letter), this may explain the difference.

If you'd like some further regression testing before merging, please let me know.  I can run the prepared example on any system you wish, including printing to a physical printer.

THausherr:
> There are failing tests on MacOS but as I can see the same tests fail on MacOS even without the fix

Just curiosity, I won't be able to do anything myself: what tests fails on MacOS?

tresf:
> > There are failing tests on MacOS but as I can see the same tests fail on MacOS even without the fix
> 
> Just curiosity, I won't be able to do anything myself: what tests fails on MacOS?

FWIW, my results on M1 / OpenJDK@11 all pass on the OP's branch `PDFBOX-5715-TRUNK`:

```
[INFO] Reactor Summary for Apache PDFBox 4.0.0-SNAPSHOT:
[INFO] 
[INFO] PDFBox parent ...................................... SUCCESS [  0.964 s]
[INFO] Apache PDFBox io ................................... SUCCESS [  4.623 s]
[INFO] Apache FontBox ..................................... SUCCESS [  4.240 s]
[INFO] Apache XmpBox ...................................... SUCCESS [  2.406 s]
[INFO] Apache PDFBox ...................................... SUCCESS [ 28.867 s]
[INFO] Apache PDFBox Debugger ............................. SUCCESS [  0.631 s]
[INFO] Apache PDFBox tools ................................ SUCCESS [ 11.235 s]
[INFO] Apache PDFBox application .......................... SUCCESS [  3.188 s]
[INFO] Apache PDFBox Debugger application ................. SUCCESS [  2.728 s]
[INFO] Apache PDFBox examples ............................. SUCCESS [ 11.809 s]
[INFO] Apache PDFBox ...................................... SUCCESS [  0.014 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:11 min
[INFO] Finished at: 2023-11-30T11:23:49-05:00
[INFO] ------------------------------------------------------------------------
```

AlexanderScherbatiy:
> Just curiosity, I won't be able to do anything myself: what tests fails on MacOS?

I am sorry. I found that I had used ssh to build and install PDFBox jars to the local maven repository and some tests failed because they couldn't initialize the graphics environment.

I checked that PDFBox tests pass with and without the fix with UI on MacOS.

THausherr:
Thank you so much for this, @AlexanderScherbatiy. Please close your PR, it has been committed. (I forgot to add the close command in the commit msg)

AlexanderScherbatiy:
The committed fix: https://github.com/apache/pdfbox/commit/73f2fff550ada9db999f7f22dca781782c5fb5bc

PR #175, opened 2024-01-04
Possible fix for Oracle's Bug JDK-8314112

viktor-merkel:
This PR fixes the NegativeArraySizeException during print when scaling causes width and height to be 0, which can lead to application freeze unexpectedly.

The sun.awt.windows.WPathGraphics class in Java contains a code segment within the drawImageToPlatform function that can lead to a NegativeArraySizeException under specific conditions. This issue arises when the scaling values in AffineTransform are set in a way that causes w and h (width and height) to have a value of 0.

"Proposed Solution:
To prevent the occurrence of the NegativeArraySizeException, a check should be introduced in the code to ensure that w and h are not 0 before proceeding with the calculation of minDpi. If either w or h is found to be 0, an appropriate error handling mechanism should be implemented to avoid division by zero."
Oracle bug submission:
https://[bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8314112](https://bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8314112)

This PR provides a fix for PDFBox that checks for a correct scale before forwarding the application flow to the problematic part of the JRE.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/rendering/AffineTransformImageValidator.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/rendering/AffineTransformImageValidator.java
@@ -1,0 +1,315 @@
+package org.apache.pdfbox.rendering;
+
+import java.awt.Graphics2D;
+import java.awt.Image;
+import java.awt.Shape;
+import java.awt.geom.AffineTransform;
+import java.awt.geom.Point2D;
+import java.awt.geom.Rectangle2D;
+import java.awt.image.BufferedImage;
+import java.awt.print.PrinterJob;
+import java.lang.reflect.Method;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+
+
+final class AffineTransformImageValidator {
+
+    private static final Log LOG = LogFactory.getLog(AffineTransformImageValidator.class);
+
+    /**
+     * For a drawing application the initial user space
+     * resolution is 72dpi.
+     */
+    private static final int DEFAULT_USER_RES = 72;
+
+    private AffineTransformImageValidator() { }
+
+    private static Class<?> pathGraphicsClass;
+
+    static {
+        try {
+            pathGraphicsClass = Class.forName("sun.print.PathGraphics");
+        } catch (ClassNotFoundException e) {
+            LOG.debug("sun.print.PathGraphics class not found");
+            pathGraphicsClass = null;
+        }
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


THausherr:
I'm wondering whether calls like `setAccessible(true)` will make trouble in newer java versions. Aren't these "unsafe"? I don't remember the details, I just remember that certain calls would be forbidden by default.

I'm also wondering whether these stunts are needed. `isBitmaskTransparency()` could be implemented without:
https://code.yawk.at/java/12/java.desktop/sun/print/PathGraphics.java#1210

Same for `getPrinterJob()`, `PrintGraphics` is a public interface.

THausherr:
I'm also wondering whether it wouldn't be easier (and faster on runtime) to just catch the NegativeArraySizeException in PageDrawer at the three places you indicated, until java fixes that bug. (despite all the work you put into that fix, sorry)

PascalSchumacher:
> I'm wondering whether calls like `setAccessible(true)` will make trouble in newer java versions. Aren't these "unsafe"? I don't remember the details, I just remember that certain calls would be forbidden by default.

Yes, on Java 16+ `reflection to access the non-public elements of java.* packages, and all elements of sun.* and other internal packages` is forbidden by default and has to be explicitly enabled, see: https://openjdk.org/jeps/396`https://openjdk.org/jeps/396 for details.

lehmi:
I agree with @THausherr the proposed patch is kind of hacky and it just solves a corner case. I'd prefer the alternative solution, catching the exception and waiting for a java fix

viktor-merkel:
@THausherr Thanks for taking a look at my pull request.

The problem with the simple "catch exceptions" solution is that it does not prevent the program from attempting to create a very large data buffer due to the division by zero, and this leads to an OutOfMemoryError of the entire application:

Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.desktop/java.awt.image.DataBufferByte.<init>(DataBufferByte.java:76)
	at java.desktop/java.awt.image.Raster.createInterleavedRaster(Raster.java:266)
	at java.desktop/java.awt.image.BufferedImage.<init>(BufferedImage.java:376)
	at java.desktop/sun.awt.windows.WPathGraphics.redrawRegion(WPathGraphics.java:1399)
	at java.desktop/sun.print.RasterPrinterJob.printPage(RasterPrinterJob.java:2320)
	at java.desktop/sun.print.RasterPrinterJob.print(RasterPrinterJob.java:1654)
        ...

THausherr:
Yeah I should have tested your program. I reverted the whole thing for a different reason, the exception doesn't happen at the place where I catch it. So you're right, the method shouldn't be called at all, i.e. the call should be prevented.

Please try to adjust your code so that it doesn't do any of the "forbidden" calls, and without jdk code.

PR #178, opened 2024-02-02
Guard Filter.decode against empty filterList

pingpingy1:
If an empty list is provided to the `Filter.decode` static method, then a `NullPointerException` is thrown just before the return statement, where `randomAccessWriteBuffer` has not been set.
The error message is thus of little help in debugging, as it gives no indication that the `filterList` argument is at fault here. This commit adds a guard to throw an `IllegalArgumentException` before the main execution to provide a more informative error trace.


THausherr:
Done, thanks!

PR #179, opened 2024-02-02
drop trailing blanks

DvonHolten:


DvonHolten:
too much on the plate - will come back with smaller bytes

THausherr:
I'm somewhat undecided about fixing trailing blanks or formatting although I have done it in the past, I prefer not to do changes in actual code because it kindof "pollutes" the history. Typos and javadoc improvements are fine.

DvonHolten:
the trailing blanks are removed automatically when saving the file. i know it creates a lot of changes, but eventually they are all gone. Considering the total size of the project, these are just the last few survivors. And they are removed only from files which received some other edits - i am not going after just the blanks

PR #180, opened 2024-02-03
Enable Native Markdown Extraction in Apache PDFBox

rawatsaurav01:
Propose the addition of native Markdown extraction support in Apache PDFBox to simplify the conversion of PDF content to Markdown, eliminating the need for intermediate HTML conversion.

**Description:**
Currently, Apache PDFBox supports HTML extraction through `PdfText2HTML`. However, this requires an extra step of converting HTML to Markdown using external tools like CopyDown. To enhance efficiency, we suggest incorporating native Markdown extraction support within Apache PDFBox.

**Sample Code Comparison:**

**Current Process:**

```java
File pdfFile = new File("sample/sample.pdf");
File mdFile = new File("sample/sample.md");
PDFText2HTML pdfText2HTML = new PDFText2HTML();
CopyDown copyDown = new CopyDown();
try (PDDocument pdDocument = Loader.loadPDF(pdfFile)) {
    Files.writeString(mdFile.toPath(), copyDown.convert(pdfText2HTML.getText(pdDocument)));
}

**Proposed Process:**

```java
File pdfFile = new File("sample/sample.pdf");
File mdFile = new File("sample/sample.md");
PDFText2Markdown pdfText2Markdown = new PDFText2Markdown();
try (PDDocument pdDocument = Loader.loadPDF(pdfFile)) {
    Files.writeString(mdFile.toPath(), pdfText2Markdown.getText(pdDocument));
}

**Benefits:**

1. **Streamlined Workflow:** Direct PDF to Markdown conversion without relying on external tools.
3. **Performance Improvement:** Reduced resource consumption, especially for large PDF files.
4. **Enhanced User Experience:** Aligns with common use cases, improving overall usability.

**Proposed Changes:**

Introduce `PDFText2Markdown` in Apache PDFBox to provide native Markdown extraction.

**Compatibility:**

Ensure backward compatibility with existing PDFBox functionalities while seamlessly adding Markdown extraction.

### DIFF
--- a/tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java
+++ b/tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java
@@ -1,33 +1,49 @@
-//package com.pdfexample;
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 package org.apache.pdfbox.tools;
 
 
-import org.apache.pdfbox.pdmodel.PDDocument;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 import org.apache.pdfbox.pdmodel.font.PDFontDescriptor;
 import org.apache.pdfbox.text.PDFTextStripper;
 import org.apache.pdfbox.text.TextPosition;
 
-import java.io.IOException;
-import java.util.*;
-
 /**
- * Wrap stripped text in simple HTML, trying to form HTML paragraphs. Paragraphs
- * broken by pages, columns, or figures are not mended.
+ * Convert PDF text to Markdown format.
+ * Each line in the PDF is converted to a corresponding Markdown paragraph.
+ * Bold and italic formatting is also applied based on font properties.
  *
- * @author John J Barton
- *
+ * @author Saurav Rawat
  */
-public class PDFText2Markdown extends PDFTextStripper{
-    private static final int INITIAL_PDF_TO_HTML_BYTES = 8192;
+public class PDFText2Markdown extends PDFTextStripper {
 
-    private final FontState fontState = new FontState();
+
+    private final org.apache.pdfbox.tools.PDFText2Markdown.FontState fontState = new org.apache.pdfbox.tools.PDFText2Markdown.FontState();
+
     /**
      * Constructor.
-     * @throws IOException If there is an error during initialization.
      */
-    public PDFText2Markdown() throws IOException
-    {
+    public PDFText2Markdown() {
         setLineSeparator(LINE_SEPARATOR);
         setParagraphStart(LINE_SEPARATOR);
         setParagraphEnd(LINE_SEPARATOR);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by THausherr** [comment]
Seems like a good idea. @lehmi do we need an ICLA for this one?

**Comment by lehmi** [comment]
@THausherr IMHO it is not needed as the class is more or less a copy of PDFText2HTML

File: tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java [hunk 1]
```diff
@@ -0,0 +1,375 @@
...
+        setArticleEnd(LINE_SEPARATOR);
+    }
+
+    @Override
+    protected void startDocument(PDDocument document) throws IOException
+    {
+        StringBuilder buf = new StringBuilder(INITIAL_PDF_TO_HTML_BYTES);
+
+        super.writeString(buf.toString());
+    }
```
lehmi: IMHO, this doesn't make sense. The used StringBuilder is empty so that nothing is written at all.

tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java
```diff
@@ -0,0 +1,375 @@
...
+        if (isLTR)
+        {
+            super.writeString(LINE_SEPARATOR);
+        }
+        else
+        {
+            super.writeString(LINE_SEPARATOR);
+        }
+    }
+
```
lehmi, line 107: Both cases are doing the same, so that the if/else can be removed

lehmi:
@rawatsaurav01 thanks for the contribution.  I've added some comments to your code and here are some more

* add a apache license header to the file
* remove/replace all HTML related comments/javadocs
* remove commented code

tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java
rawatsaurav01: @lehmi I have removed line as per the suggestion.

tools/src/main/java/org/apache/pdfbox/tools/PDFText2Markdown.java
rawatsaurav01: @lehmi Please check I have updated the source code.

rawatsaurav01:
@lehmi @THausherr Thanks a lot for your encouraging comments . I have encorporated the changes. Please review and let me know if some changes required.

THausherr (changes requested):
Please put back Axels name at line 164, he should still be credited as your co-author IMHO.

rawatsaurav01:
@THausherr I have changed the author name to Axle . Can you please check it now.

THausherr:
Thank you @rawatsaurav01 I committed to the trunk with some minor changes. I'll commit to 3.0 and 2.0 later.

rawatsaurav01:
Thank you @THausherr .

THausherr:
Sonar found a missing break in appendEscaped() which is no big deal, but then I realized that some of the code isn't needed. At the same time some needed code should be added, i.e. handling of special characters like "*". See also 
https://shadocs.netlify.app/markdown/escaping_character/

rawatsaurav01:
Hi @THausherr I have done some changes in the code. Previously it was having issue in handling special characters like *,#,+,-. But now the issue is resolved . Actually it takes * as dot and # as a line so in the new edit I have added  a backslash('\') for the characters which are giving errors.
 I have added a screenshot of https://markdownlivepreview.com/ after converting a pdf (having special charaters) to markdown. 
Screenshot 2024-02-17 232046

THausherr:
I don't see any additional commit, either add it or post the code change here. I assume it's in `appendEscaped()`.

rawatsaurav01:
@THausherr I have added and pushed the code ,but it is not showing maybe the pr is closed.Here is the updated code.

```java
private static void appendEscaped(StringBuilder builder, char character) {

        switch (character) {
            case '*':
                builder.append("\\" + character);
                break;
            case '+':
                builder.append("\\" + character);
                break;
            case '-':
                builder.append("\\" + character);
                break;
            case '#':
                builder.append("\\" + character);
                break;
            case 34:
                builder.append("\"");
                break;
            case 38:
                builder.append("&");
                break;
            case 39:
                builder.append("\'");
                break;
            case 60:
                builder.append("<");
                break;
            case 62:
                builder.append(">");
                break;
            case 178:
                builder.append("<sup>2</sup>");
                break;
            case 179:
                builder.append("<sup>3</sup>");
                break;

            default:
                builder.append(character);
                break;
        }
    }

```

THausherr:
Thank you, I've simplified this and committed it. If there isn't any further input then I'll commit to versions 2 and 3 in a few days.

rawatsaurav01:
Thank you @THausherr for your acknowledgement.

THausherr:
All done in https://issues.apache.org/jira/browse/PDFBOX-5768 .

johmue:
@THausherr @rawatsaurav01 This is really helpful, especially now with LLMs trained on MarkDown formatted documents. Would it be a big effort to expose this option in the PDFBox App with a dedicated SubCommand i.e. `export:md` or with `export:text -md` (similar to -html)?

THausherr:
Yes sounds like a good idea, I'm gonna think about it tomorrow.

THausherr:
Done  https://issues.apache.org/jira/browse/PDFBOX-5895 .
It will soon be available at
https://repository.apache.org/content/groups/snapshots/org/apache/pdfbox/pdfbox-app/2.0.33-SNAPSHOT/
https://repository.apache.org/content/groups/snapshots/org/apache/pdfbox/pdfbox-app/3.0.4-SNAPSHOT/
Please test it.

johmue:
It works and creates good outcomes. Thanks for putting the effort and making it easily available in almost no time.

PR #181, opened 2024-02-05
update some JavaDoc

DvonHolten:
- update JavaDoc for class to show if this table is required in a TrueType-font.
- strip trailing blanks

1.) i see some inconsistencies wrt short/int. The read() method reads the fields as short/int as required, but some classes just convert it to int, whereas other classes keep the 'short'-type, even in getters. What to do?
2.) the read() methods set the initialized-flag when successfully finished. However, this is sticky. When there is a chance, that objects of these xxxTable-classes are reused, the initialized flag should be cleared at method-entry and set at method-end. What to do?


THausherr:
Something went wrong here, this PR has 207 commits, but not by you?!

DvonHolten:
tja...
i try again - shall the pullrequests go to the trunk or to some branch?

THausherr:
Always against the trunk

PR #182, opened 2024-02-05
update some JavaDoc in Font-table classes

DvonHolten:
the JavaDoc fixes are preparation for some bug-hunt wrt OpenType-font.


THausherr:
Thanks, I applied your changes in the trunk and in 3.0 with some small modifications: only one IOException in the javadoc; some grammar changes. If I'm right about these, then it proves you were not using an AI, which is a good thing.

PR #183, opened 2024-02-06
Use computed font height as font size

GooDer:
Use computed font height as font size in case that form field has dynamic font size (0) and its value is empty so width can't be used

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/AppearanceGeneratorHelper.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/AppearanceGeneratorHelper.java
@@ -914,6 +914,11 @@ class AppearanceGeneratorHelper
                 }
 
                 float heightBasedFontSize = contentRect.getHeight() / height * yScalingFactor;
+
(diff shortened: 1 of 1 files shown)


GooDer:
Hello,

I have encounter problem during flattening of one of our PDF. I think that main reason of failure is wrong usage of form field in the source PDF but I can’t change it (I think it was generated by some tool). I can’t also shared the problematic PDF since it contains sensitive data.

During processing of flattening there is a method to calculate size of font for these dynamic form field and in this case it leads to division by zero float and ending as negative infinite.

It is edge case but still I think it would be good to handle it.

Element displayed in PDFBox debugger:
image

Values and place where problem happens in debugger:
image


I think that only height values should be taken into consideration in this case.

Used PDFBox 3.0.1, java 17

THausherr:
Thanks, I was able to reproduce this problem by using a very small rectangle width for the form field. I'm gonna fix this tomorrow.

THausherr:
I committed a modified code that only checks for infinity because 1) it's more clear, 2) the method still returns a negative font size if the vertical size is negative, however this has no bad effect. I also added a test. Thanks for the report!

PR #184, opened 2024-02-08
some ToDo s to make things better

DvonHolten:
- add a ToDo proposal avoid recompiling regex (twice!) on every use
- add a ToDo proposal to save a lot (?) of memory
- remove some trailing blanks

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocument.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocument.java
@@ -347,9 +347,6 @@ public class PDDocument implements Closeable
             throw new IllegalStateException("Cannot sign an empty document");
         }
 
-        int startIndex = Math.min(Math.max(options.getPage(), 0), pageCount - 1);
-        PDPage page = pageTree.get(startIndex);
-
         // Get the AcroForm from the Root-Dictionary and append the annotation
         PDDocumentCatalog catalog = getDocumentCatalog();
         PDAcroForm acroForm = catalog.getAcroForm(null);
@@ -378,12 +375,15 @@ public class PDDocument implements Closeable
             acroForm.getCOSObject().setItem(COSName.FIELDS, new COSArray());
         }
         PDAnnotationWidget firstWidget;
+        PDPage page;
         if (signatureField == null)
         {
             signatureField = new PDSignatureField(acroForm);
             // append the signature object
             signatureField.setValue(sigObject);
             firstWidget = signatureField.getWidgets().get(0);
+            int startIndex = Math.min(Math.max(options.getPage(), 0), pageCount - 1);
+            page = pageTree.get(startIndex);
             // backward linking
             firstWidget.setPage(page);
         }
@@ -391,6 +391,7 @@ public class PDDocument implements Closeable
         {
             firstWidget = signatureField.getWidgets().get(0);
             sigObject.getCOSObject().setNeedToBeUpdated(true);
+            page = null;
         }
 
         // TODO This "overwrites" the settings of the original signature field which might not be intended by the user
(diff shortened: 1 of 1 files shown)


THausherr:
Feel free to add a commit about a compiled constant regex. I don't have an opinion on the other topic.

DvonHolten:
ok, regex coming.
the flag-thing should have some investigation on how many objects are created. is it 10 or 10000?

xzel23:
Nice find.

For the flag thing, how about just changing isValid to this? 
```
    public boolean isValid()
    {
        return value != Long.MAX_VALUE && value != Long.MIN_VALUE;
    }
```

This would also make the constructor and where it is called a little bit shorter because there'd be no need to pass the `valid`parameter.

And the `STATIC` array could just be initialised in a static block, IMHO that wouldn't cost too much memory, and get(long) would be much simpler:

```
    public static COSInteger get(long val)
    {
        if (LOW <= val && val <= HIGH)
        {
            return STATIC[(int) val - LOW];
        }
        return new COSInteger(val, true);
    }
```

`getInvalid(boolean maxValue)` can also be removed and we use just
```
    protected static final COSInteger OUT_OF_RANGE_MAX = new COSInteger(Long.MAX_VALUE);
    protected static final COSInteger OUT_OF_RANGE_MIN = new COSInteger(Long.MIN_VALUE);
```

UPDATE: there would of course be a little semantic change because in the current implementation it's possible to `get(Long.MAX_VALUE)` to obtain a valid `COSInteger` with that value. But that looks rather like a bug in the current implementation to me because we'd have two instances where `equals()`returns `true` while one is valid and the other is not. Someone with more insight into the PDF spec has to decide if the current semantics though weird are mandated by the spec or not.

xzel23:
@THausherr What do you think about the suggested change? I just "bought" the PDF spec yesterday (you have to buy it, but the price is 0 USD), and I couldn't find anything about COInteger to require Long.MIN_VALUE and Long.MAX_VALUE to map to valid COSInteger values.

@DvonHolten if Tilman and/or some other maintainer agrees to the proposed change, do you want to make the flag related change yourself or should I create a PR for that?

THausherr:
I'm not deep enough into the COSInteger code, it is related to PDFBOX-5176 and any change (except removing `getInvalid()` because of https://twitter.com/unclebobmartin/status/1114137614377005057 ) should be checked with the file from there and with the file from the earlier issue. PDFBOX-5176 was created after running a regression test on 250000 PDF files.

DvonHolten:
i can do it - check, where the invalid things are coming from and where they are checked. if all the Long-values must be available, a subclass may do it.
and i have a closer look at those 250000 pdf -files in PDFBOX-5176. i assume, they are just created to check the parser - not to render them?

THausherr:
You don't have to bother with the 250000 files, this is our regression test corpus and we run these once on a dedicated server before doing a release. Just test the files from PDFBOX-5176 and the one from the related issue. I mentioned the 250000 files to show how unlikely this is.

fontbox/src/main/java/org/apache/fontbox/ttf/GlyphSubstitutionTable.java
```diff
@@ -75,6 +76,13 @@ public class GlyphSubstitutionTable extends TTFTable
 
     private GsubData gsubData;
 
+    /**
+     * The regex represents 4 'word characters' [a-zA-Z_0-9],
+     * see {@link java.util.regex.ASCII#WORD}<br>
+     * Note: the ' '-character is not matched!
+     */
+    static final Pattern wordPattern = Pattern.compile( "\\w{4}" );
```
xzel23: The name should be all uppercase for constants (static final). It should also be private.

xzel23: What do you think about making this a predicate?

```
    private static final Predicate<String> IS_FOUR_LETTERS = Pattern.compile( "\\w{4}" ).asMatchPredicate();
```

Then the test would be `IS_FOUR_LETTERS.test(featureTags[i])`.

fontbox/src/main/java/org/apache/fontbox/ttf/GlyphSubstitutionTable.java
```diff
@@ -799,10 +806,11 @@ public int getSubstitution(int gid, String[] scriptTags, List<String> enabledFea
     }
 
     /**
-     * For a substitute-gid (obtained from {@link #getSubstitution(int, String[], List)}), retrieve the original gid.
-     *
-     * Only gids previously substituted by this instance can be un-substituted. If you are trying to unsubstitute before
-     * you substitute, something is wrong.
+     * For a substitute-gid (obtained from {@link #getSubstitution(int, String[], List)}),
+     * retrieve the original gid.
+     * <br>
```
xzel23: Usually `<p>` is used in Javadoc to start a new paragraph.

xzel23 (commented):
Note that I am not an official project member, so I just left some comments. I think the name of the matcher should be changed and made private, the comment about using a Predicate is just a suggestion that the maintainers may or may not like.

fontbox/src/main/java/org/apache/fontbox/ttf/GlyphSubstitutionTable.java
THausherr: yeah, can be done for the trunk (needs jdk11). I'll use the previous suggestion for 2.0 and 3.0.

xzel23 (commented):
Looks good to me now. Thank you!

THausherr:
I've now committed the javadoc changes and the pattern change.

PR #186, opened 2024-02-21
suppress warnings for octals, omit trailing blanks

DvonHolten:
- suppress warnings on octal constants in Encoding-tables
- ToDo potential error: missing leading 0 in MacOSRomanEncoding.java
- drop trailing blanks

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
@@ -25,7 +25,7 @@ import org.apache.pdfbox.pdmodel.common.COSObjectable;
 
 /**
  * A PostScript encoding vector, maps character codes to glyph names.
- * 
+ *
  * @author Ben Litchfield
  */
 public abstract class Encoding implements COSObjectable
@@ -75,7 +75,7 @@ public abstract class Encoding implements COSObjectable
 
     /**
      * Returns an unmodifiable view of the code -&gt; name mapping.
-     * 
+     *
      * @return the code -&gt; name map
      */
     public Map<Integer, String> getCodeToNameMap()
@@ -97,9 +97,9 @@ public abstract class Encoding implements COSObjectable
     /**
      * This will add a character encoding. An already existing mapping is preserved when creating
      * the reverse mapping. Should only be used during construction of the class.
-     * 
+     *
      * @see #overwrite(int, String)
-     * 
+     *
      * @param code character code
      * @param name PostScript glyph name
      */
@@ -136,7 +136,7 @@ public abstract class Encoding implements COSObjectable
 
     /**
      * Determines if the encoding has a mapping for the given name value.
-     * 
+     *
      * @param name PostScript glyph name
      * @return true if the encoding has a mapping for the given name value
      */
@@ -147,10 +147,10 @@ public abstract class Encoding implements COSObjectable
 
     /**
      * Determines if the encoding has a mapping for the given code value.
-     * 
+     *
      * @param code character code
      * @return if the encoding has a mapping for the given code value
-     * 
+     *
      */
     public boolean contains(int code)
     {
@@ -159,23 +159,18 @@ public abstract class Encoding implements COSObjectable
 
     /**
      * This will take a character code and get the name from the code.
-     * 
+     *
      * @param code character code
-     * @return PostScript glyph name
+     * @return PostScript glyph name, or ".notdef" when not found.
      */
     public String getName(int code)
     {
-       String name = codeToName.get(code);
-       if (name != null)
-       {
-          return name;
-       }
-       return ".notdef";
+       return codeToName.getOrDefault(code, ".notdef");
     }
 
     /**
      * Returns the name of this encoding.
-     * 
+     *
      * @return the name of the encoding
      */
     public abstract String getEncodingName();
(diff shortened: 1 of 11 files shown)


pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacOSRomanEncoding.java
```diff
@@ -27,7 +27,9 @@ public class MacOSRomanEncoding extends MacRomanEncoding
     /**
      * Table of octal character codes and their corresponding names
      * on top of {@link MacRomanEncoding}.
+     * ToDo are these really octals?? leading 0 seems to be missing
```
xzel23, line 30: Good catch!

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
```diff
@@ -159,23 +159,18 @@ public boolean contains(int code)
...
      */
     public String getName(int code)
     {
-       String name = codeToName.get(code);
-       if (name != null)
-       {
-          return name;
-       }
-       return ".notdef";
+       return codeToName.getOrDefault(code, ".notdef");
```
xzel23, line 168: getOrDefault() is available from Java 8 on, ok for PDFBox  3+.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacExpertEncoding.java
```diff
@@ -27,6 +27,7 @@ public class MacExpertEncoding extends Encoding
     /**
      * Table of octal character codes and their corresponding names.
      */
+    @SuppressWarnings("OctalInteger")
```
xzel23, line 30: I think that's IntelliJ specific and might in turn produce a warning in other IDEs.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacRomanEncoding.java
```diff
@@ -29,6 +29,7 @@ public class MacRomanEncoding extends Encoding
     /**
      * Table of octal character codes and their corresponding names.
      */
+    @SuppressWarnings("OctalInteger")
```
xzel23, line 32: same as above

xzel23 (commented):
Note: I'm not  a PDFBox committer, so I have no authority here.

- the numbers in MacRomanEncoding should be octal - compare the values with this [table](https://en.wikipedia.org/wiki/Mac_OS_Roman) in Wikipedia.
- I think the @SuppressWarnings annotations are specific to IntelliJ and not standard
- getOrDefault() should be added for PDFBox 3+ (it's not available in Java 6 for PDFBox 2)

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
DvonHolten: getOrDefault() is used in 2 other locations as well

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacOSRomanEncoding.java
lehmi: Thanks, I've added the leading "0", see PDFBOX-5773

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacExpertEncoding.java
lehmi: I won't add that as it is IntelliJ specific

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
lehmi: I've added the proposed change as well, see PDFBOX-5660

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/MacRomanEncoding.java
DvonHolten: there are other SuppressWarnings (like "squid:S106" or "SpellCheckingInspection"). There are just a few warnings defined in Java, so that annotation mostly adresses 'other tools'.

lehmi: We are using sonar and I guess it was added to address that one

THausherr: Yes, either that one or another tool we were using.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/encoding/Encoding.java
xzel23: > getOrDefault() is used in 2 other locations as well

Yes, I just made that comment to remind whoever integrates this change doesn't pull it into PDFBox 2. Many fixes and improvements are backported to the old release that still supports Java 6.

DvonHolten:
octals are fixed

PR #188, opened 2024-04-01
Fix containers inefficiencies in Type1CharString.java, COSWriterCompressionPool.java, and ShadedTriangle.java

cinsttool:
Hi,

We find that there exist container inefficiencies in Type1CharString.java, COSWriterCompressionPool.java, and ShadedTriangle.java.

In particular, at line 301 of Type1CharString.java, the object `list` copies the `numbers` and removes the last two elements for calculation purposes before adding the result, which is then returned by the method `handleType1Command`. Apart from this scenario, method `handleType1Command` returns `Collections.emptyList()`. However, at lines 154 and 155, after the the invocation of `handleType1Command`, the `numbers` object is cleared, and the `results` object are copied to the `numbers` object using the `addAll` method. This process involves redundant copying and clearing operations.

For COSWriterCompressionPool.java, at the line 56, the created `List` object `allDirectObjects` is only used for `contains` and `add` operations within the `addElements` method. Such usage scenarios are more suitable for a `Set` container with lower complexity.

For ShadedTriangle.java, according to the prurpose of method calcDeg, there is too expensive to create objects and pass them to a `Set` to obtain the count of unique elements and then returning the count.

We discovered the above containers inefficiencies by our tool cinst. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

### DIFF
--- a/fontbox/src/main/java/org/apache/fontbox/cff/Type1CharString.java
+++ b/fontbox/src/main/java/org/apache/fontbox/cff/Type1CharString.java
@@ -168,6 +168,7 @@ public class Type1CharString
             // indicates an invalid charstring
             LOG.warn("Unknown charstring command in glyph {} of font {}", glyphName, fontName);
             numbers.clear();
+            return;
         }
         switch(type1KeyWord)
         {
--- a/fontbox/src/main/java/org/apache/fontbox/cff/Type1CharString.java
+++ b/fontbox/src/main/java/org/apache/fontbox/cff/Type1CharString.java
@@ -150,9 +150,7 @@ public class Type1CharString
         type1Sequence.forEach(obj -> {
             if (obj instanceof CharStringCommand)
             {
-                List<Number> results = handleType1Command(numbers, (CharStringCommand) obj);
-                numbers.clear();
-                numbers.addAll(results);
+                handleType1Command(numbers, (CharStringCommand) obj);
             }
             else
             {
@@ -161,7 +159,7 @@ public class Type1CharString
         });
     }
 
-    private List<Number> handleType1Command(List<Number> numbers, CharStringCommand command)
+    private void handleType1Command(List<Number> numbers, CharStringCommand command)
     {
         commandCount++;
         Type1KeyWord type1KeyWord = command.getType1KeyWord();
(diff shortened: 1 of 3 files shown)


THausherr:
I'm undecided about the first one, I like the second, I don't like the third one because the code looks much more complex. I assume you're hard-coding the usage of the 3-set.

cinsttool:
> I'm undecided about the first one, I like the second, I don't like the third one because the code looks much more complex. I assume you're hard-coding the usage of the 3-set.

Yes, we only evaluate container usage patterns from a performance perspective. For example, regarding the third one, we consider that using a container solely to obtain a size instead of persistently storing and managing objects may be too costly. However, the modification we provided may indeed be less concise and clear compared to the original implementation.

cinsttool:
> > I'm undecided about the first one, I like the second, I don't like the third one because the code looks much more complex. I assume you're hard-coding the usage of the 3-set.
> 
> Yes, we only evaluate container usage patterns from a performance perspective. For example, regarding the third one, we consider that using a container solely to obtain a size instead of persistently storing and managing objects may be too costly. However, the modification we provided may indeed be less concise and clear compared to the original implementation.

Maybe we can set initial capacity of the set to 3 that can save memory consumption while maintain the usage of the 3-set as the original implementation.

fontbox/src/main/java/org/apache/fontbox/cff/Type1CharString.java
```diff
@@ -161,15 +159,15 @@ private void render()
...
+    private void handleType1Command(List<Number> numbers, CharStringCommand command)
     {
         commandCount++;
         Type1KeyWord type1KeyWord = command.getType1KeyWord();
         if (type1KeyWord == null)
         {
             // indicates an invalid charstring
             LOG.warn("Unknown charstring command in glyph {} of font {}", glyphName, fontName);
-            return Collections.emptyList();
+            numbers.clear();
```
lehmi, line 170: The immediate return is missing, so that the switch runs into the default case and throws an exception

cinsttool: Sorry for the mistake! We push this patch.

lehmi:
@cinsttool  thanks for the contribution. I've applied the remaining changes for Type1CharString

The changes are limited to the trunk and 3.x as it would have included a breaking change in 2.x

lehmi:
All tests passed without any error, but the rendering is broken. I'm going to have a look

lehmi:
My bad, I missed one line. Finally everything is fine now

PR #192, opened 2024-05-02
Use assert throws

jens-kaiser:
user assertThrows and other assertions from JUnit 5, switch to jdk 21


THausherr:
Hello, I just committed about half of your changes. Thank you!

PR #194, opened 2024-05-16
Change error log level to warning.

GooDer:
It is some auto correction and PDF will be processed even with these problems so I want to avoid error alerts to be triggered.

User mail list doesn't work for me somehow so I am proposing it this way.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDPageTree.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDPageTree.java
@@ -197,7 +197,7 @@ public class PDPageTree implements COSObjectable, Iterable<PDPage>
                     if (set.contains(kid))
                     {
                         // PDFBOX-5009, PDFBOX-3953: prevent stack overflow with malformed PDFs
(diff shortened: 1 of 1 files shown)


THausherr:
These are nasty problems. The PDF is definitively messed up and it should be checked with the sender whether it is correct. Just because you get the same content than with Adobe doesn't mean everything is OK.

GooDer:
Hi @THausherr sorry for late reply.

I agree that PDF is not in good shape, but unfortunately it is some input from users / companies which we need to process even if there are some problems.

Logging that error is problem for our alerting system, which is triggered for every error logged.

I can see three possibilities, how to handle this:

1. as proposed in PR, change log level to warning. PDF is still processed even that this error occure
2. gather errors in some list and allow user of the library to decide what to do in case of error
3. switch of error logging for that class, which could be dangerous
4. allow to provide some configuration to parser/library in which level this should be logged (with error as default)

Please suggest what would be the best choice in this case.

Thanks you

THausherr:
I like none; however with log4j you can limit output in specific classes.

GooDer:
Ok, I will go with option 3 as you suggest.

PR #195, opened 2024-05-20
PDFBOX-5823: remove regexp matcher to reduce memory utilisation

jonathansp:
PDAbstractContentStream uses StringUtil.PATTERN_SPACE regexp to evaluate if a word has a space in it (https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDAbstractContentStream.java#L1624)

For large documents ~800 pages and small string sequences (like a regular word), it causes a memory overhead (see attached), due to the several extra allocations. I've replaced the regexp for space and \t using word.contains, and since it's a O ( 1 ) operation that does not require extra allocations, memory used has been reduced.

https://issues.apache.org/jira/browse/PDFBOX-5823

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDAbstractContentStream.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDAbstractContentStream.java
@@ -1621,7 +1621,8 @@ abstract class PDAbstractContentStream implements Closeable
 
         for (String word : words)
         {
-            if (StringUtil.PATTERN_SPACE.matcher(word).matches())
+            if (word == null) continue;
+            if (word.length() == 1 && word.isBlank())
             {
                 out.write(font.encode(word));
             }


DvonHolten:
but what is the purpose of this? if you just want/need to know, if some string contains a space ' ', the fastest and cheapest way is to use indexOf( ' ' ) >= 0.
if you need to check for two (or very few) possible characters, use some index.Of(), ordered by expected frequency.
Patterns cant be more efficient than that - for these simple cases. 
If you need to match for something like 'whitespace', use a compiled pattern.
Dont try to build your own pattern matcher!

jonathansp:
@DvonHolten patterns work fine but also cause a **massive** memory overhead while creating large documents. The pattern is compiled, however it will be slower than `String.indexOf` anyways. In order to cover whitespaces, we'd need to call `indexOf()` a few times, providing different chars `(" ", \t, \n, \r, \f ...)`, which `isblank()` is for, regardless its char table, ordered by the most common frequency.

Please see PDFBOX-5823 and attached benchmarks.

jonathansp:
Solved in https://svn.apache.org/viewvc/pdfbox/branches/3.0/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDAbstractContentStream.java?view=markup&pathrev=1917878

PR #196, opened 2024-05-21
PDFBOX-5824: allow map threshold to be set as system property

jonathansp:
[COSDictionary.MAP_THRESHOLD](https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/cos/COSDictionary.java#L54) controls which Map class is used to optimize memory usage. By default, a SmallMap is used. However, if the number of items in a COSDictionary reaches the MAP_THRESHOLD value (hardcoded to 1,000), the references [are copied ](https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/cos/COSDictionary.java#L208)to a LinkedHashMap.

For larger documents, where the COSDictionary is expected to be substantial bigger than this limit, this copying occurs frequently. Additionally, [SmallMap.keySet is not efficient](https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/util/SmallMap.java#L281).

Some properties already supported in this library are:

* pdftextstripper.indent
* pdftextstripper.drop
* org.apache.pdfbox.filter.deflatelevel

https://issues.apache.org/jira/browse/PDFBOX-5824

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/cos/COSDictionary.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/cos/COSDictionary.java
@@ -50,13 +50,14 @@ public class COSDictionary extends COSBase implements COSUpdateInfo
      */
     private static final Logger LOG = LogManager.getLogger(COSDictionary.class);
 
+    private static final String SYSPROP_SMALLMAP_THRESHOLD = "org.apache.pdfbox.cosdictionary.smallmap.threshold";
     private static final String PATH_SEPARATOR = "/";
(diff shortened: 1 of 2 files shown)


jonathansp:
As per discussion, SmallMap has been deprecated. Solved in https://svn.apache.org/viewvc?view=revision&revision=1918260 (https://github.com/apache/pdfbox/commit/c894e5d46bcdc06cb9b6e525cbbf78036d90b811)

PR #197, opened 2024-08-22
PDFBOX-5872: Support imageio-openjpeg for JPEG2000 reading

Googulator:
The imageio-openjpeg doesn't implement raster reading, but PDFBox no longer relies on raster support, so just removing that check makes imageio-openjpeg work with PDFBox when the required native library is installed.

However, when the native library is not installed, imageio-openjpeg indicates that it's not usable by returning null as the image reader. This needs to be checked to avoid NPEs merely from this library being installed without its native component.


THausherr:
Now you have 5 files changed and a method that is harder to read.
Please watch this: https://www.youtube.com/watch?v=2Q9GRPxqCAk

Googulator:
Making the non-raster version private would mean leaving JBIG2 with an unnecessary restriction to readers with raster support only.

Now, unlike JPEG2000, I'm not aware of any problematic JBIG2 reader, but semantically it's still more correct to accept non-raster-capable readers.

PR #198, opened 2024-12-19
Add method to merge PDF from InputStream source

aaronhagopian:
Introduced a new `addSource` method that accepts an `InputStream` as a parameter, enabling merging of PDFs directly from streams keeping the developer from having to understand the RandomAccessRead interfaces and use a common way of accessing file data.

Ideally this is merged with the 3.0 branch as well.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
@@ -16,10 +16,7 @@
  */
 package org.apache.pdfbox.multipdf;
 
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
```diff
@@ -16,10 +16,7 @@
  */
 package org.apache.pdfbox.multipdf;
 
-import java.io.File;
```
aaronhagopian, line 19: Just realizing this now, intellij did that automatically

THausherr:
I'd prefer to improve the javadoc. Also for `Loader.loadPDF()` which has the same problem.

lehmi:
I fully support @THausherr. We are using the RandomAccessRead interfaces to give the user maximum flexibility w.r.t. the used access method and possible optimization

PR #199, opened 2025-01-02
Updated formatting for better visibility

esivakumar26:


THausherr:
Why change it? It's currently bold so your change rather makes it less visible.

PR #202, opened 2025-02-26
Small cleanup to GSUB-related code in PDAbstractContentStream

gredler:
- Add missing JavaDoc
- Simplify a few variable names, in some cases allowing wrapped lines to fit on one line
- Avoid unnecessary extra cast
- Pre-size ByteArrayOutputStream to size which should usually be enough to fit all encoded glyphs
- Remove unnecessary special cases from word loop
- Avoid need to throw IOExceptions by using ByteArrayOutputStream-specific writeBytes method
- Avoid unnecessary codePointCount( ) call to get number of code points
- Simplify IllegalStateException error message to match similar messages elsewhere in the codebase

Happy to tweak the changes, if they look mostly good but there are one or two things that need to be done differently.


gredler:
Closing -- changes which made sense already incorporated separately by @THausherr . Thanks!

PR #203, opened 2025-02-27
PDFBOX-5230: force zero-width chars to be invisible

gredler:
First pass.

NOTE: one of the tests does not currently pass, looks like something to do with the char / glyph ID / CID mapping?

### DIFF
--- a/fontbox/src/main/java/org/apache/fontbox/ttf/GlyfSimpleDescript.java
+++ b/fontbox/src/main/java/org/apache/fontbox/ttf/GlyfSimpleDescript.java
@@ -51,7 +51,7 @@ public class GlyfSimpleDescript extends GlyfDescript
         super(numberOfContours);
 
         /*
-         * https://developer.apple.com/fonts/TTRefMan/RM06/Chap6glyf.html
+         * https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6glyf.html
          * "If a glyph has zero contours, it need not have any glyph data." set the pointCount to zero to initialize
          * attributes and avoid nullpointer but maybe there shouldn't have GlyphDescript in the GlyphData?
          */
--- a/fontbox/src/main/java/org/apache/fontbox/ttf/TTFSubsetter.java
+++ b/fontbox/src/main/java/org/apache/fontbox/ttf/TTFSubsetter.java
@@ -53,7 +53,7 @@ public final class TTFSubsetter
 {
     private static final Logger LOG = LogManager.getLogger(TTFSubsetter.class);
     
-    private static final byte[] PAD_BUF = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+    private static final byte[] PAD_BUF = { 0, 0, 0, 0 };
 
     private static final TimeZone TIMEZONE_UTC = TimeZone.getTimeZone("UTC"); // clone before using
 
@@ -631,6 +631,13 @@ public final class TTFSubsetter
                             isResult);
                 }
 
+                // glyphs with no outlines have an empty entry in the 'glyf' table, with a
+                // corresponding 'loca' table entry with length = 0
+                if (invisibleGlyphIds.contains(gid))
+                {
+                    continue;
+                }
+
                 byte[] buf = new byte[(int)length];
                 isResult = is.read(buf);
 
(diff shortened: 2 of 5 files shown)


THausherr:
Here's your font: 
[testEmbeddedFontWithZeroWidthChars.zip](https://github.com/user-attachments/files/19027325/testEmbeddedFontWithZeroWidthChars.zip)

Here's what DTL OTMaster 3.7 thinks of it:
grafik

And this is the glyf table in that font:
grafik

Number of points 36610 ?! Note that the "A" glyph has only 17.

`font.getPath('\u200C').getBounds2D()` returns `java.awt.geom.Rectangle2D$Double[x=205.0,y=0.0,w=1125.0,h=1409.0]`

THausherr:
I found one bug but fixing it doesn't change anything for THIS file.

https://learn.microsoft.com/en-us/typography/opentype/spec/glyf
"If a glyph has zero contours, no additional glyph data beyond the header is required." this is misleading because the next sentence is "A glyph with zero contours may have additional data, however; in particular, it may have instructions that operate on phantom points" so you'd need to add two 0 bytes for instructionLength. But doing it doesn't change anything because the 4-byte alignment adds two bytes.

THausherr:
path content:

```moveTo(205.0, 1409.0);
lineTo(1330.0, 1409.0);
lineTo(1330.0, 0.0);
lineTo(205.0, 0.0);
lineTo(205.0, 1409.0);
closePath();
moveTo(281.0, 1333.0);
lineTo(281.0, 76.0);
lineTo(1254.0, 76.0);
lineTo(1254.0, 1333.0);
lineTo(281.0, 1333.0);
closePath();
```

That's the rectangle of GID 0 ?! However getPath() uses the code from the PDF (see the javadoc!), not the unicode. The code in the PDF is 0x7d9 and then the test passes.

I now wonder if it is a bug in DTL OTMaster. I also looked in FontForge and that one didn't have anything weird. However FontForge is a pain to use and maybe I missed something.

THausherr:
This test passes:
```
@Test
void testEmbeddedFontWithZeroWidthChars() throws IOException
{
    String text = "AAA\u200CBBB";
    ByteArrayOutputStream baos = new ByteArrayOutputStream();
    int code;
    try (PDDocument document = new PDDocument())
    {
        PDPage page = new PDPage();
        document.addPage(page);
        InputStream input = PDFont.class.getResourceAsStream(
                "/org/apache/pdfbox/resources/ttf/LiberationSans-Regular.ttf");
        PDType0Font font = PDType0Font.load(document, input);
        byte[] encoded = font.encode('\u200C');
        assertEquals(2, encoded.length);
        code = ((encoded[0] & 0xFF) << 8) | (encoded[1] & 0xFF);
        try (PDPageContentStream stream = new PDPageContentStream(document, page))
        {
            stream.beginText();
            stream.setFont(font, 20);
            stream.newLineAtOffset(50, 600);
            stream.showText(text);
            stream.endText();
        }
        document.save(baos);
    }
    try (PDDocument document = Loader.loadPDF(baos.toByteArray()))
    {
        // verify that the text still contains zero-width characters
        PDFTextStripper stripper = new PDFTextStripper();
        String extractedText = stripper.getText(document);
        assertEquals(text, extractedText.trim());

        // verify that the zero-width characters are invisible
        PDPage page = document.getPage(0);
        PDResources resources = page.getResources();
        Iterable< COSName > fontNames = resources.getFontNames();
        COSName fontName = fontNames.iterator().next();
        PDType0Font font = (PDType0Font) resources.getFont(fontName);
        assertEquals(0, font.getWidth(code));
        assertEquals(0, font.getWidthFromFont(code));
        assertTrue(font.getPath(code).getBounds2D().isEmpty());
    }
}
```

gredler:
:+1: Very cool, I'm in the processing of upgrading my OTMaster license, but will hopefully have it in a day or two and will check things on my end and update the PR per your feedback!

gredler:
I was also getting strange results (one point, when none should be present). I used OTM to have a look at the space glyph in the Liberation Sans Regular font, and after looking at the 'loca' table I realized that we don't need to write any data to the glyf table for these glyphs. This results in a zero-length entry in the 'loca' table, and things seem to work correctly. I'm going to tidy things up over the next few days and update the PR so that you can re-review.

gredler:
@THausherr OK, second pass committed. I think this one is looking good. Let me know if you see anything that needs to be changed!

THausherr:
Yes this looks good, DTL OTMaster is no longer complaining, so we don't need to know whether the version I was using had a bug or not. I'll commit it soon (but probably not today).

THausherr:
Please close your PR, I forgot to include the command when committing to the trunk.

gredler:
@THausherr Done -- thanks!

PR #208, opened 2025-07-04
Add option to skip corrupt PDFs in PDFMergerUtility with improved exception handling

SwethaMuthuvel:
## What This PR Does

This pull request improves the robustness and debuggability of `PDFMergerUtility` by:

1. **Adding a `skipCorruptFiles` flag**  
   - Allows users to skip unreadable or corrupt PDF files during merge.  
   - Default behavior remains unchanged (i.e., throws on error).

2. **Wrapping `IOException` with source context**  
   - Converts vague errors like:  
     ```
     IOException: Could not parse object stream
     ```  
     into more useful messages like:  
     ```
     IOException: Failed to load PDF from source: /path/to/file.pdf
     ```  
   - Helps identify exactly which file failed.

3. **Applied consistently in both merge modes**  
   - `optimizedMergeDocuments(...)`  
   - `legacyMergeDocuments(...)`  
   - Added warning logs when skipping files.

---

## Why This Helps

- **Improves debuggability** — pinpoints which file caused the failure.
- **Makes batch operations resilient** — avoids total failure from one bad input.
- **Scales better** — suitable for bulk merging scenarios.
- **Does not break existing behavior** — opt-in via `setSkipCorruptFiles(true)`.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/multipdf/PDFMergerUtility.java
@@ -102,6 +102,24 @@ public class PDFMergerUtility
     private DocumentMergeMode documentMergeMode = DocumentMergeMode.PDFBOX_LEGACY_MODE;
     private AcroFormMergeMode acroFormMergeMode = AcroFormMergeMode.PDFBOX_LEGACY_MODE;
     
+    private boolean skipCorruptFiles = false;
+
+    /**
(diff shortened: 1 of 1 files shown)


lehmi:
Please reformat the code first using our formatter rules to make it easier to evaluate your proposed changes

THausherr:
I'm wondering what the use case of this change would be. Wouldn't the target file be worthless if parts of the source is missing?

Is this for a school / university project, or is this part of an AI training / evaluation?

PR #210, opened 2025-07-12
PDFBOX-6032: add configurable default image factory

kimilg:
- This PR addresses [PDFBOX-6032](https://issues.apache.org/jira/browse/PDFBOX-6032) by introducing a new overload of PDImageXObject#createFromByteArray that accepts a user-supplied defaultFactory parameter for customizable image conversion behavior.
- The goal is to allow users to provide a defaultFactory for custom image conversion. This factory is used directly for formats like BMP and GIF, and acts as a fallback for PNG and TIFF when the optimized conversion path fails and would otherwise fall back to LosslessFactory.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CustomFactory.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CustomFactory.java
@@ -5,6 +5,6 @@ import java.io.IOException;
 import org.apache.pdfbox.pdmodel.PDDocument;
 
 @FunctionalInterface
-public interface DefaultFactory {
+public interface CustomFactory {
 	PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray) throws IOException;
 }
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CustomFactory.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CustomFactory.java
@@ -4,7 +4,19 @@ import java.io.IOException;
 
 import org.apache.pdfbox.pdmodel.PDDocument;
 
+/**
+ * A functional interface that allows users to define a custom strategy for converting
+ * image data as a byte array into a {@link PDImageXObject}.
+ */
 @FunctionalInterface
 public interface CustomFactory {
+	/**
+	 * Creates a {@link PDImageXObject} from the given image byte array and document context.
+	 *
+	 * @param document the document that shall use this PDImageXObject.
+	 * @param byteArray the image data as a byte array
+	 * @return a PDImageXObject.
+	 * @throws IOException if there is an error when creating the PDImageXObject.
+	 */
 	PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray) throws IOException;
 }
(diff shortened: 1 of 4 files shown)


pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/PDImageXObject.java
```diff
@@ -383,6 +383,69 @@ public static PDImageXObject createFromByteArray(PDDocument document, byte[] byt
...
+     * @param name name of image file for exception messages, can be null.
+     * @param defaultFactory optional factory used to handle BMP, GIF, or fallback cases 
+     *                       (e.g., for PNG or TIFF). If {@code null}, this method delegates to
+     *                       {@link #createFromByteArray(PDDocument, byte[], String)}.
+     * @return a PDImageXObject.
+     * @throws IOException if there is an error when reading the file or creating the
+     * PDImageXObject.
+     * @throws IllegalArgumentException if the image type is not supported.
+     */
+    public static PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray, String name, DefaultFactory defaultFactory) throws IOException
```
rototor: Here I'm a bit confused, also by the naming stuff.

I would rather extend the existing method with a parameter. I think "DefaultFactory" is a bad name. I.e., the way I would implement this would be 

```
// Method with signature of the existing method is just forwarding to the other  new overload
public static PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray, String name)
{
return createFromByteArray(document,byteArray,name, null);
}

// Extend the existing method with the one parameter
public static PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray, String name, DefaultFactory defaultFactory) throws {
  // .. logic like it is now BUT instead of just encoding this lossless we first check if we have a *fallback*

if( defaultFactory != null ) 
  return defaultFactory.create...(...)
// otherwise just do the PNGEncoder Lossless stuff like it is now.
}
```

And therefor I think defaultFactory is bad name. It should be a fallback.

lehmi: I've proposed DefaultFactory and I'm not really happy with that name. 

Fallback won't be correct as it isn't a fallback in all cases. It is for png and tiff which can't be processed but it isn't for bmp and gif.

How about CustomFactory?

kimilg: Thanks for the feedback.  
I’ve removed the duplicate logic, and the code feels more compact and cleaner now.
Naming the functional interface is tricky - a precise name gets too long, but a short one feels too vague.
I renamed it to CustomFactory because it more clearly reflects its intended usage than DefaultFactory.

pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CustomFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.pdfbox.pdmodel.graphics.image;
+
+import java.io.IOException;
+
+import org.apache.pdfbox.pdmodel.PDDocument;
+
+@FunctionalInterface
+public interface CustomFactory {
```
rototor, line 12: Some JavaDoc to describe what this interface is about would be nice.

But other than that, the changes look fine for me.

kimilg: Right, I forgot to add it earlier. I've now added a JavaDoc comment to this class.

THausherr:
Thanks for the PR; I forgot the magic words in the commit message, please close the PR yourself.

PR #214, opened 2025-08-17
avoid type conversion in PdPage.getContentsForStreamParsing method

valerybokov:


THausherr:
Thanks, but can't we go even further by just doing "if (COSName.FLATE_DECODE.equals(filter))" ?

valerybokov:
I understood. I will rewrite it.

PR #215, opened 2025-08-17
remove useless check in NumberFormatUtil.formatPositiveNumber method

valerybokov:


THausherr:
Thanks, but if "remaining > Integer.MAX_VALUE" is true then "remaining > 0" is also true. Thus "(!omitTrailingZeros || remaining > 0)" would always be true regardless of the value of "omitTrailingZeros".

I also asked ChatGPT and it did the same that you did, but when I started to argue, it agreed with me. (However I have the impression that ChatGPT tries to agree with the user)

valerybokov:
I not used ChatGPT but what I see? The check what works from left to right. We have 2 branches:
1) the remaining variable is bigger then Integer.MAX_VALUE and  it is bigger then 0. So we able to exclude second check
2) the remaining variable is bigger then Integer.MAX_VALUE and omitTrailingZeros is false

So, we really able to exclude the condition '&& (!omitTrailingZeros || remaining > 0)'.

I left it as is because I thought here is some bug. Okay. Can I do another pull request with this fix to remove the '&& (!omitTrailingZeros || remaining > 0)' condition?

PR #230, opened 2025-08-30
remove useless final operator in Filter class methods

valerybokov:


THausherr:
How is it "useless"? The class itself isn't final.

valerybokov:
This is not class this is method (findImageReader, findRasterReader). Static method cannot be overriden.

THausherr:
It's more complex. What you mean is that when having the same method in the derived class, this isn't real overriding, it's just a method with the same signature. I tested this by adding this method
```
    public static ImageReader findImageReader(String formatName, String errorCause)
            throws MissingImageReaderException
    {
        return null;
    }
```
in LZWFilter.java. Without the change, my IDE (netbeans) shows an error. With the change, the error goes away. So the "final" does have the intended effect, preventing users of adding these two methods in a derived class.

valerybokov:
Understood.

PR #232, opened 2025-08-30
improve performance for PDComplexFileSpecification.getFilename method

valerybokov:


THausherr:
I prefer to keep this one the way it is, IMHO it looks nicer currently. The performance win is minimal and this method isn't used by the majority of the users.

valerybokov:
okay

PR #237, opened 2025-08-30
improve performance for PDNumberTreeNode.getNumbers method

valerybokov:


xzel23:
You pass the size as one argument to the HashMap constructor. This does not exactly do what you expect it to do, I.e., avoid rehashing. The default load factor of 0.75 is used. What you would really want to use is HashMap.newHashMap(int size), but that was only introduced in java 19. The proper workaround for older java versions would be to use new HashMap(n * 4 / 3).

THausherr:
Ouch. Ok, I'll revert these recent changes and won't do these anymore. Adding the calculation won't make the code more readable.

PR #244, opened 2025-09-06
improve performance for PDStructureTreeRoot.getClassMap

valerybokov:


THausherr:
Although I assume there is no breaking test, this might not be helpful in case a class map is retrieved to add elements because the new map is immutable.

PR #247, opened 2025-09-06
improve write operations for RC4Cipher

valerybokov:
The idea is to reduce the time of write operations by writing an array into the output stream rather than individual bytes where possible. This approach has pos and cons. The pos: reducing operations time. The cons: this array should not be very big or we will get an OutOfMemoryException. According the tests, the maximum length of that array is 1024. If you think so, I can rewrite the write(byte[], int, int, OutputStream) method by setting a maximum lenght of the array (buffer, a line 146). It will be something like in the method write( InputStream, OutputStream).

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/encryption/RC4Cipher.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/encryption/RC4Cipher.java
@@ -95,6 +95,14 @@ class RC4Cipher
         data[ secondIndex ] = tmp;
     }
 
+    private int encrypt(byte aByte) {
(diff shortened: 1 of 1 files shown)


THausherr:
I took the time to review it today and it looks ok and useful (I'm wondering whether we should have "& 0xFF" in the method that assigns an int result to a byte? But the build or the rendering tests should fail if it is needed) but I'll look at it again tomorrow or later. I also think that in another step we can get rid of methods that aren't used. Only SecurityHandler.java uses it.

valerybokov:
I think XOR operation between two bytes can't return anything larger or smaller than a byte.

PR #255, opened 2025-09-19
Add option to implement custom `ICOSVisitor` behaviour for `COSObject`

jp-db:
`COSObject` overrides the `COSBase#accept` method to visit its inner object. 

However, this approach limits the ability of an `ICOSVisitor` superclass to define a custom behaviour when a `COSObject` is visited. 

This patch adds the method `ICOSVisitor#visitFromObject` to provide for flexibility for such cases.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/cos/COSObject.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/cos/COSObject.java
@@ -160,15 +160,7 @@ public class COSObject extends COSBase implements COSUpdateInfo
     @Override
     public void accept( ICOSVisitor visitor ) throws IOException
     {
-        COSBase object = getObject();
-        if (object != null)
-        {
-            object.accept(visitor);
-        }
-        else
-        {
-            COSNull.NULL.accept(visitor);
-        }
+        visitor.visitFromObject(this);
     }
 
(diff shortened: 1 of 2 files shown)


THausherr:
I think the proposed patch would be rather for 3.0 and 2.0 to preserve compatibility, while an ideal patch for the trunk would have an abstract method for `visitFromObject` and your proposed "default" code would go into a method in `COSWriter`, do you agree or disagree? (You don't have to change the patch, I'd adjust it if you agree with my understanding)

jp-db:
I agree with your suggestion

THausherr:
The 3.0 commit is at https://github.com/apache/pdfbox/commit/a0b20ac80091dd97aebc76b86536caf2b40aecff

I did not make a change for 2.0, I had forgotten that this is for jdk8 which does not have default methods in interfaces.

A snapshot build will be at 
https://repository.apache.org/content/groups/snapshots/org/apache/pdfbox/pdfbox-app/3.0.6-SNAPSHOT/

The PDFBox issue is at 
https://issues.apache.org/jira/browse/PDFBOX-6073

If you have further improvements / bug reports and made a user application in the past that was refused (sorry!), please make a new one, just mention something that we know it's you.

PR #260, opened 2025-09-21
improve performance for PDMeshBasedShadingType.readPatch method

valerybokov:


valerybokov:
In previous version were 2 calls: getBitsPerCoordinate and getBitsPerComponent. I extracted the variables from them. Then you changed a bitsPerComponent variable (line 222) to bitsPerCoordinate variable.

THausherr:
Ouch, thanks for noticing. That's because I did it manually instead of using the patch. I'll fix it tomorrow.

PR #261, opened 2025-09-21
remove unneeded command from testFromAnnots3891ValidateFont

valerybokov:


THausherr:
Thanks, I've replaced it with the assumed intention of the author.

PR #262, opened 2025-09-23
improve performance for PDAbstractAppearanceHandler.handleBorderBox method

valerybokov:


THausherr:
I just reverted this one due to differences in the rendering tests. `getRectangle()` calls `annotation.getRectangle()` so the old version of the code gets the rectangle modified by `annotation.setRectangle()`. One might still do something (it's called twice before the change), but this is really a trivial method that is likely to be optimized by java.

PR #265, opened 2025-09-27
improve performance for AcroFormOrphanWidgetsProcessor.ensureFontResources method

valerybokov:


THausherr:
Thanks, but this one I'll fix differently because it's only used in debug log mode.

PR #270, opened 2025-09-27
add capacity for list in-TrueTypeEmbedder.subset method

valerybokov:


xzel23:
Maybe just use List<String> tables = List.of("head", "hhea", ...);

valerybokov:
> Maybe just use List tables = List.of("head", "hhea", ...);

Yes. One of the solutions.

xzel23:
What I mean is: the effect of your change at runtime is negligible: the old version initializes the internal array reference with a pre-allocated shared Object[] instance of length 0 (since Java 9, I think before Java 9, a new empty array was allocated every time) and the first add operation instantly allocates a new array of size 10 (which is by chance the default capacity of an ArrayList). Since the empty array is a shared static instance, even the number of array allocations remains the same (this would not be true for size > 10). But the new version is slightly less readable than the original version, and should the list ever have to be expanded or elements removed, the size has to be changed manually too.

Compared to that, using List.of(...) reduces the number of method calls by 10, returns a list that is allocated with the correct size right from the start no matter how many elements you pass (no reallocations are necessary), and the code is both more readable and easier to maintain (less code and no manual keeping track of the array size).

It should also perform slightly better. But if this really was a performance bottleneck, the tables list should be converted to a private static final constant, this only minimally increases the heap size because all strings contained in the table are contained in the string pool anyway as long as the program runs, so only the list overhead would be added, while at the same time the code size and number of memory allocations during runtime would be reduced.

All in all, I see lots of small PRs you made with the title "improve performance of ...". But in most cases the performance impact is near zero. That being said, your PRs generally really improve the code quality - so you could maybe name them "improve code quality" - which is something that is at least as important as performance, so keep it up - extracting common expressions into temporary variables often improves readability and maintainability (like a change needs to be made only once instead of several times when the same expression is used multiple times).

valerybokov:
Thank you for the clarification regarding List.of. I used the ArrayList constructor with the initial capacity because I didn't want to make significant changes to the code. I see that inside the List.of method, the listFromTrustedArray method is called, and inside it, there's a loop with a null check and a check for the collection size. Although, I don't argue that List.of could be faster. If Tilman Hausherr prefers the List.of approach, I'll rework it.

xzel23:
Yes, important differences between the list returned by List.of() and new ArrayList():
- List.of() does not allow null values
- the list returned by List.of() is immutable, I.e., you can't add or remove elements.
- the null checks you see when inserting gives the hotspot compiler the opportunity to omit null checks when retrieving elements as (at least if I recall correctly) the JVM "knows" this implementation can never return null elements.

But all these do not matter in this case.

IMHO, if you are sure about the changes you make, run the tests and think about any possible side effects, there's little reason not to make slightly bigger changes if the result is preferable: better readability/maintainability, performance, less allocations etc.

If your code does not get through, you probably will learn something new that you did not think about, which also is a good thing.

THausherr:
IMHO List.of() doesn't add readability. If this is a constant then it means that classes that extend that class won't be able to add tables in the future. I'd add the 10 (despite that this is the default) so that nobody feels the need to optimize this in the future.

lehmi:
The usage of the list in question is limited to TrueTypeEmbedder.subset(). It is one parameter for the constructor of TTFSubsetter, which itself is final and can't be extended. 

Looks like the list can be considered as constant, so that we might create a private static final instance using Arrays.asList or List.of

Using a constant will avoid creating the list every time TrueTypeEmbedder.subset is called.

THausherr:
You're right, I had missed that TrueTypeEmbedder is package local. Thus yes a constant is fine.

lehmi:
@THausherr you've committed 2 different implementations. I guess due to the fact that List.of isn't available in java versions < 9. Maybe Arrays.asList is the better solution as it is available in all relevant java version. AFAIK there is just one difference Array.asList creates a mutable list.
However, none of the solutions uses a constant. The list is still created every time the method is called

THausherr:
Yes, thanks, I missed that comment. I'll fix it.

lehmi:
Thanks for the fast response and sorry for bothering you again

You may replace the static code block with Arrays.asList("head", "hhea" ....)

PR #274, opened 2025-09-27
fix memory leak for RandomAccessReadBuffer(InputStream)

valerybokov:
1 I think we should close this stream especially if we got it by this way (CMapParser.getExternalCMap):

InputStream is = getClass().getResourceAsStream(name);

2 Possible optimisation for future:
We often use ByteArrayInputStream as parameter of the constructor RandomAccessReadBuffer(InputStream). The ByteArrayInputStream is the array wrapper. There is no need to create List<ByteBuffer> for this case.

### DIFF
--- a/io/src/main/java/org/apache/pdfbox/io/RandomAccessReadBuffer.java
+++ b/io/src/main/java/org/apache/pdfbox/io/RandomAccessReadBuffer.java
@@ -110,19 +110,19 @@ public class RandomAccessReadBuffer implements RandomAccessRead
         int remainingBytes = chunkSize;
         int offset = 0;
         byte[] eofCheck = new byte[1];
-        while (remainingBytes > 0 &&
-                (bytesRead = input.read(currentBuffer.array(), offset, remainingBytes)) > -1)
-        {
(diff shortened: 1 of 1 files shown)


lehmi:
The caller of the constructor should close the resource as we don't know if the input stream maybe reused

Saying that, we should close the stream in getExternalCMap

valerybokov:
> The caller of the constructor should close the resource as we don't know if the input stream maybe reused
> 
> Saying that, we should close the stream in getExternalCMap

This is one of the possible solutions. Yes, it can be closed outside. Will you rewrite it or me? There are several tests what have the same memory leak.

lehmi:
I'll have a look. It seems there are other occurrences as well

lehmi:
Done, see https://issues.apache.org/jira/browse/PDFBOX-6079

@valerybokov Thanks for the pointer

valerybokov:
Your idea is based on current solution when RandomAccessReadBuffer reads all the data inside its constructor. That is why you able to create a new instance of the RandomAccessReadBuffer and close the stream immediately.

PR #286, opened 2025-10-11
fix typo in NonSeekableRandomAccessReadInputStream.rewind

valerybokov:


valerybokov:
About another code. There is a strange solution in PDFMarkedContentExtractor.processTextPosition (216). The list is empty, so condition is always true.

THausherr:
Thank you, I'll think about this one. Probably delete it completely. It has been like this for 15 years. I assume this code is modified from the text stripper to create a simple solution for marked content extraction.

PR #287, opened 2025-10-11
cancel heavy close operations if the stream is already closed

valerybokov:


io/src/main/java/org/apache/pdfbox/io/RandomAccessReadMemoryMappedFile.java
```diff
@@ -110,6 +110,7 @@ public void close() throws IOException
         if (fileChannel != null)
         {
             fileChannel.close();
+            fileChannel = null;
```
lehmi, line 113: This isn't needed. The garbage collector takes care of final members once the class itself is no longer referenced

valerybokov: The idea isn't to nullify references, but to avoid calling the close and other methods multiple times, since they are often synchronized.

lehmi: OK, that is a reasonable point. IMHO the close operation should be moved to the if clause of the mappedByteBuffer

PR #288, opened 2025-10-11
speed up for DataInputRandomAccessRead.readBytes method

valerybokov:


THausherr:
The length check with available() isn't needed because available() does a similar calculation than the one just above the change.

valerybokov:
> The length check with available() isn't needed because available() does a similar calculation than the one just above the change.

Yes, i see it.

valerybokov:
https://github.com/apache/pdfbox/pull/289

PR #289, opened 2025-10-11
speed up for DataInputRandomAccessRead.readBytes method v2

valerybokov:


THausherr:
Sadly there are 5 tests that fail (also with the previous version).

CFFParserTest.testCharset() fails because 2108 bytes are read, but the length to read is 53597.

This isn't a bug, the javadoc doesn't promise that it will read everything possible, or that there is an EOF if it reads less. We'd need a readFully() method.

valerybokov:
> Sadly there are 5 tests that fail (also with the previous version).
> 
> CFFParserTest.testCharset() fails because 2108 bytes are read, but the length to read is 53597.
> 
> This isn't a bug, the javadoc doesn't promise that it will read everything possible, or that there is an EOF if it reads less. We'd need a readFully() method.

understood

THausherr:
I wrote some code that loops until the buffer is full. However I will commit that one after the release next week.

THausherr:
Fixed in https://issues.apache.org/jira/browse/PDFBOX-6083

PR #290, opened 2025-10-13
close PDFStreamParser

valerybokov:
The PDFStreamParser has a close method. Should it be closed outside or the close method should be protected (private)?


lehmi:
The parser is closed in org.apache.pdfbox.pdfparser.PDFStreamParser.parseNextToken() once the end is reached.

I'm not sure if I like that behaviour as it isn't transparent to the user .... maybe we should refactor that first.

valerybokov:
> The parser is closed in org.apache.pdfbox.pdfparser.PDFStreamParser.parseNextToken() once the end is reached.
> 
> I'm not sure if I like that behaviour as it isn't transparent to the user .... maybe we should refactor that first.

So why isn't it protected (PDFStreamParser.close)?

PR #291, opened 2025-10-13
extract setCOSValue in PDField.importFDF

valerybokov:
This path excludes mentions of the PDTerminalField and PDChoice types from PDField to make the code more object-oriented. Additionally, the child types will get less checks inside an importFDF(FDFField) method.


THausherr:
I prefer to keep this as it is. The change splits the complexity into a different complexity at different places.

PR #294, opened 2025-10-25
close stream in CreatePDFA.main method

valerybokov:


valerybokov:
Another question. 
PDIndexed class has 3 constructors. The constructor without parameters not calls a initRgbColorTable method. When I try to do something like this 

PDIndexed pd = new PDIndexed();
pd.toRGBImage(...);

then I get NullPointerException because rgbColorTable is null.

THausherr:
One might set it with the COS API but I haven't seen any of it. Our code has one single case where such an object is created not by reading from a PDF and that is in PNGConverter and that one doesn't use the no parameter constructor. So I'll deprecate it and remove it in the trunk until somebody complains.
https://issues.apache.org/jira/browse/PDFBOX-6089

PR #303, opened 2025-10-30
add ChangeTextColor example to modify text color in PDF documents

KrishVij:
### PDFBOX-3.6: Add ChangeTextColor example

This pull request adds a new example class **`ChangeTextColor`** to the PDFBox examples module  
(`org.apache.pdfbox.examples.pdmodel`).

The example demonstrates how to modify the **text color** in an existing PDF by parsing its content streams,  
identifying text sections, and injecting RGB color operators (`rg` and `RG`) immediately after  
the `BT` (Begin Text) operator.

---

#### 🔍What the example shows

- Shows how to use **`PDFStreamParser`** to access and modify PDF operators.  
- Demonstrates rebuilding content streams with **`ContentStreamWriter`**.  
- Converts standard RGB values (0–255) into PDF-compatible color space (0.0–1.0).  
- Uses only public PDFBox API classes — no private/internal APIs.
Before:
Screenshot 2025-10-30 220526
After:
Screenshot 2025-10-30 220700

PS: Both are screenshot of a resume of mine

---

### DIFF
--- a/examples/src/main/java/org/apache/pdfbox/examples/pdmodel/ChangeTextColor.java
+++ b/examples/src/main/java/org/apache/pdfbox/examples/pdmodel/ChangeTextColor.java
@@ -1,0 +1,123 @@
+package org.apache.pdfbox.examples.pdmodel;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.pdfbox.Loader;
+import org.apache.pdfbox.cos.COSFloat;
+import org.apache.pdfbox.pdfparser.PDFStreamParser;
+import org.apache.pdfbox.pdfwriter.ContentStreamWriter;
+import org.apache.pdfbox.pdmodel.PDDocument;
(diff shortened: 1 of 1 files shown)


KrishVij:
Can anybody at least tell me  was my contribution not upto the mark it will mean a lot to me.

THausherr:
Sorry for the late reaction. I'm kinda undecided. The problem is that this will work only for very simple cases, where the text is in the top content stream.

KrishVij:
Greetings Mr/Ms Maintainer

Pls dont be sorry! I can understand that being  a maintainer of a big open
source project like this will be of course time consuming and hard. Pls
decide and let me know whether this will be merged and on the flip note if
you can tell me where to study more about the content streams I will really
appreciate it as I did read the spec but due to my current experience and
current expertise it was a bit hard to understand  these complex concepts

THausherr:
Look at the examples subproject and then look at the RemoveAllText.java example, this processes form objects and patterns too. But be aware that PDF is so complex that there will always be PDFs that will fail to be changed because there are so many different ways to do anything.

PR #313, opened 2025-11-02
close streams in-TestCreateSignature

valerybokov:


THausherr:
Thank you; somehow the patch couldn't be used on my system but I think I got it all, I also refactored / removed some code.

PR #316, opened 2025-11-08
simplify several methods in COSArray

valerybokov:


THausherr:
While this seemingly simplifies the code, it also reassigns a parameter, which would be flagged by the Sonar code analyzer.

xzel23:
I have to say that IMHO the new code is much simpler and easier to read and maintain and would suggest simply accepting this in the Sonar web interface so that it won't show up any more in future runs of the analyzer. But it is a valid point and it's up to the project to decide whether marking issues as accepted in Sonar is an acceptable solution.

Another solution would be to simply define a new variable and assign it in the body of the if-statement. I think that would still make the code easier to read.

THausherr:
While there is no project policy on this, I'd prefer the second solution, and I agree with the last sentence.
The very last "if" isn't needed, this makes the code harder to read and in most cases the new value will be different than the old onne.

(Please do not send PRs on existing code about this parameter issue)

valerybokov:
https://github.com/apache/pdfbox/pull/321

PR #317, opened 2025-11-08
set capacity for list in GlyphRenderer.calculatePath method

valerybokov:


THausherr:
there are more entries added after that loop, shouldn't it be +3 ? +1 because of <=, +2 because at most 2 entries are added.

valerybokov:
> there are more entries added after that loop, shouldn't it be +3 ? +1 because of <=, +2 because at most 2 entries are added.

According to current benchmarks, the list's size is always equal to its capacity. And yes, we're adding one element, but it could be two. There's no test that uses code that adds two elements to a list. So, we can write 3 instead of 2. If we write 3, sometimes the list's capacity will be larger than its size (more memory). Or we can use 2, which will be slightly slower in some cases because we need to increase the capacity.

Apparently, it's a minor issue. If necessary, I'll give it a 3.

THausherr:
Yes 3 please.

valerybokov:
https://github.com/apache/pdfbox/pull/320

PR #325, opened 2025-11-11
improve parsing speed of COSFloat

valerybokov:


THausherr:
This one I don't want to change because that is only called in some error cases (see the comments) and it makes the code less readable because the regexp would be elsewhere.

PR #338, opened 2025-11-15
close streams in-LosslessFactoryTest

valerybokov:


valerybokov:
Another question:

PDLineAppearanceHandler.generateNormalAppearance.
144: contents can be null
174 -> PDFont.encode (333) -> NPE

THausherr:
Thanks I changed that one, but after doing it I realized there is no NPE risk because of `if (annotation.hasCaption() && !contents.isEmpty())`.

PR #340, opened 2025-11-18
close streams in TestCOSIncrement

valerybokov:


THausherr:
Thanks but both are closed when done (see parameter description in javadoc).

valerybokov:
> Thanks but both are closed when done (see parameter description in javadoc).

Okay. I see where a stream to load font is closed.
I see there is written for document.saveIncremental method the comment "It will be closed when done".  But I do not see where this stream is closed in fact.

valerybokov:
I checked all cases for PDDocument.saveIncremental method. For each case the stream was closed outside.

THausherr:
Thank you, I see that this was changed in https://issues.apache.org/jira/browse/PDFBOX-4321 , however the javadoc and the test weren't all fixed, I'll do that.

PR #341, opened 2025-11-18
close stream in TestExifXmp

valerybokov:
Another question.
Still don't need to close the stream for the Loader.loadXFDF(String) and Loader.loadXFDF(File file) methods? Okay, the Loader.loadXFDF(String) method can be simplified by creating a File object instead of a stream, meaning it can be called inside Loader.loadXFDF(File file).


valerybokov:
```
    public static FDFDocument loadXFDF(String filename) throws IOException
    {
        return Loader.loadXFDF(new File(filename));
    }
```

THausherr:
Thank you, I've fixed these (hopefully).

PR #345, opened 2025-11-20
close stream in Rendering.renderPDFSpec

valerybokov:


THausherr:
Thanks, I changed that it uses File directly. I assume this is a mistake because the benchmark is about rendering, not about writing the file.

PR #352, opened 2025-11-25
extract variables in PDLineAppearanceHandler.generateNormalAppearance method

valerybokov:


weihuanwan:
Hey team,

We’ve been working on advanced table recognition and structured content extraction based on PDFTextStripper. However, in version 3.0.6, the writeLine() method is declared private, which makes it impossible to customize the line-level extraction logic.

Because of this restriction, we currently have no choice but to copy a large amount of code from the original PDFTextStripper implementation just to override internal behavior. This significantly increases maintenance cost and reduces the value of relying on PDFBox as a stable dependency.

Would it be possible for the next release to change writeLine() (and related internal hooks) to protected or public, so developers can extend the class cleanly without copying internal source?

This change would make custom text-layout logic—especially table parsing—much easier to implement, and it would greatly reduce the need to fork or replicate code from PDFBox.

        Thanks for considering this improvement.

THausherr:
@192902649 please don't use this as a forum. Use the users mailing list at https://lists.apache.org/list.html?users@pdfbox.apache.org and don't forget to subscribe before you post.

PR #358, opened 2025-11-29
improve parsing speed in PDFStreamParser

valerybokov:


pdfbox/src/main/java/org/apache/pdfbox/pdfparser/PDFStreamParser.java
```diff
@@ -46,6 +47,7 @@ public class PDFStreamParser extends COSParser
      */
     private static final Logger LOG = LogManager.getLogger(PDFStreamParser.class);
 
+    private static final Pattern pattern = Pattern.compile("^\\d*\\.?\\d*$");
```
xzel23: As a constant, this should be all uppercase. I also suggest to use a more descriptive name like PATTERN_NUMBER.

And I f I don't overlook anything, the pattern is only used for testing if some string matches, so it would IMHO be even more straightforward to use a predicate:

```
Predicate<String> IS_NUMBER = Pattern.compile("\\d*\\.\\d*").asMatchPredicate();
```

And then later:
```
if (IS_NUMBER.test(...))
```

PR #363, opened 2025-12-06
delete unneeded conversion in COSParser.checkPagesDictionary

valerybokov:


THausherr:
I have a few of these, either because of Sonar warnings or IDE warnings. They don't hurt.

valerybokov:
> I have a few of these, either because of Sonar warnings or IDE warnings. They don't hurt.

Okay

PR #370, opened 2025-12-06
fix typo in PDICCBased.setRangeForComponent method

valerybokov:


valerybokov:
Hi, @THausherr! This is another question. I cannot say is this bug or not. So, I wrote this information.
I found one detail in the COSDictionary.addAll method.
I compared it with other methods like clear, setItem, removeItem. And the call "getUpdateState().update();" always exists in them. But not in addAll.
Also, I want to say that the addAll method calls the COSDictionary constructor, and maybe we don't need to call "getUpdateState().update();" in this case.

THausherr:
It looks suspicious. IMHO this should either be modified or documented. Please open a ticket in JIRA or post on the users mailing list for this type of topics.

PR #371, opened 2025-12-06
Fix condition where auto raster dpi is set to zero

mattcoz:
If the auto raster dpi is zero, the code fails when creating the buffered image with the following error:

java.lang.IllegalArgumentException: Width (0) and height (0) cannot be <= 0

This change prevents that. The value of dpi here is always equal to rasterDpi when originally set to RASTERIZE_DPI_AUTO.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/printing/PDFPrintable.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/printing/PDFPrintable.java
@@ -267,7 +267,7 @@ public final class PDFPrintable implements Printable
             // rasterize to bitmap (optional)
             Graphics2D printerGraphics = null;
             BufferedImage image = null;
-            if (dpi > 0 || dpi == RASTERIZE_DPI_AUTO)
+            if (rasterDpi > 0)
             {
                 LOG.debug("dpi set to {}", rasterDpi);
                 float dpiScale = rasterDpi / 72;


THausherr:
Could you please show an example how to reproduce the problem? From my understanding, if -1 is passed, a new `rasterDpi` is calculated. The second "if" line (the one you want to change) makes sure that a new scale is calculated.

THausherr:
I see I misunderstood your patch yesterday, now I think I got it - however I'd like to know how the calculated auto raster dpi would ever be zero. This can only happen if `graphics2D.getTransform().getScaleX()` is zero?!

mattcoz:
Yeah, that seems to be the case, although I check `getMatrix().getScalex()` right before I print and it's not zero. So I looked at the code for `PDFPageable` and noticed it mentioning that Java has trouble with landscape pages, and that happens to be the case with the document I'm trying to print. I change it to use a standard portrait paper size and it prints fine. So it seems that it could be related to [PDFBOX-2922](https://issues.apache.org/jira/browse/PDFBOX-2922).

THausherr:
I've added some logging and found the actual cause when I tested it, which is that scaleX is 0 when a page is rotated. I should have known better, because this is a well known flaw that we have encountered before, and it has now been fixed as well by using `new Matrix(graphics2D.getTransform()).getScalingFactorX()`. Thank you for finding this. You can close the PR.

PR #381, opened 2025-12-14
refactor-PDFTextStripper.writeLine method

valerybokov:
The normalize method returns an instance of the LinkedList. We use that list inside a writeLine method. The writeLine method uses the list for a loop. The previous version of the loop was slow for a list because each element was retrieved through the get method.

So, there are two ways to fix it:
1 use the foreach statement instead of the for statement
2 use ArrayList instead of LinkedList


THausherr:
I looked at the history and the LinkedList was introduced 15 years ago in PDFBOX-521 and I can't see why. Likely a habit by that developer.

valerybokov:
> I looked at the history and the LinkedList was introduced 15 years ago in PDFBOX-521 and I can't see why. Likely a habit by that developer.

I checked at all the uses of the LinkedList in the library and only this case made me suspicious.

PR #390, opened 2026-01-05
Fix typo in PDLinkAppearanceHandler.getLineWidth method

valerybokov:


valerybokov:
It looks like I added 2 different changes in one PR.

PR #400, opened 2026-01-22
set initial capacity for list in GsubWorkerForDflt.applyGsubFeature method

valerybokov:
actually this is the minimum capacity because I don't know what length this chunk is


THausherr:
Same comment as for the latin worker.

PR #401, opened 2026-01-22
set initial capacity for list in GsubWorkerForLatin.applyGsubFeature method

valerybokov:
actually this is the minimum capacity because I don't know what length this chunk is


THausherr:
Then we shouldn't change it (and the others) at all. Here's an example with the latin test:

tokens (1): [[393, 367, 258, 410, 296, 381, 396, 373]]
22.01.2026 20:09:13 INFO  gsub.GsubWorkerForLatin:118 - originalGlyphs: [393, 367, 258, 410, 296, 381, 396, 373], gsubProcessedGlyphs: [393, 367, 258, 410, 296, 381, 396, 373]

So tokens size is 1, result list is 8. This is a very complex part of the code. I was able to understand it 2 years ago but that is gone.

PR #403, opened 2026-01-24
set initial capacity for list in GsubWorkerForTamil.getbeforeRephGlyphIds-and-getRephGlyphIds method

valerybokov:


valerybokov:
A little fix. The getbeforeRephGlyphIds method uses another array.

PR #405, opened 2026-01-24
fix comment for TTFParser.parseTableHeadersa method

valerybokov:


THausherr:
Thank you, you're correct in flagging this, but I'll make a different change (and correct more). It's still a 100% font stream input, however it only parses the headers so it is faster. This was introduced in https://issues.apache.org/jira/browse/PDFBOX-5847 .

valerybokov:
> Thank you, you're correct in flagging this, but I'll make a different change (and correct more). It's still a 100% font stream input, however it only parses the headers so it is faster. This was introduced in https://issues.apache.org/jira/browse/PDFBOX-5847 .

Okay. Not a problem.

PR #413, opened 2026-02-03
close stream for CCITTFactoryTest.testCreateFromRandomAccessMulti method

valerybokov:
On my machine, the ImageReader is com.sun.imageio.plugins.tiff.TIFFImageReader. And it disposes anything. The method is empty.
So we need to additionally close this image stream.


valerybokov:
Another question. Sometimes we use javax.imageio.ImageIO.read method. Documentation of this method says that it can return null.

THausherr:
Yes it can return null, I see this isn't checked in a few non-test methods. However if this returns null then it is likely a configuration error.

PR #416, opened 2026-02-08
close documents and graphics in LosslessFactoryTest

valerybokov:


THausherr:
Thanks, however I ignored most, because `doWritePDF()` closes the input. Maybe this isn't elegant coding and I'm to blame for that, but it reduced code and it's documented 😂

PR #424, opened 2026-02-21
extract additional methods in SecurityHandler to reduce checks number

valerybokov:
I've extracted several methods because the decryptDictionary(COSDictionary, long, long) method has type checks (old code, line 616), and the decrypt(COSBase, long, long) method has the same type checks. So, there is double type check in the code. The new implementation of the decryptDictionary method calls overloads of the decrypt method for each type (COSString, COSArray, COSDictionary) to avoid double type check.

Perhaps all of these new methods should be made public (or all of them private). You're the one in charge.


THausherr:
Around 616 the old code was calling the general method. The new code is calling the specific method, but now it will no longer execute the code part that checks whether that part has already been decrypted before?!

valerybokov:
> Around 616 the old code was calling the general method. The new code is calling the specific method, but now it will no longer execute the code part that checks whether that part has already been decrypted before?!

Thank you! I've fixed the bug.

PR #427, opened 2026-03-21
Improve directory boundary check in ExtractEmbeddedFiles example

MarkLee131:
Improve directory boundary check in ExtractEmbeddedFiles example


yurii20041:
Hi Team,

It seems this change wasn’t committed and CVE is still being reported by BD. Could you please provide some updates?  Also, will version 3.0.8 be released soon?

Thanks!

PR #437, opened 2026-04-08
PDFBOX-6189 : Safely validate dimension to avoid OOM 

subbudvk:


THausherr:
thank you; please make two small improvements:
1) add the current maximum in the exception
2) check the sysprop for zero or negative; if that happens, ignore it (like you already ignore bugus values)

subbudvk:
@THausherr updated. Please check

PR #443, opened 2026-04-16
reset graphics state in PDFPrintable.print method

valerybokov:


THausherr:
You're focusing on the border painting, but the transform is also changed previously elsewhere.

THausherr:
Why check both? I assume with "cast" you mean "printerGraphics = (Graphics2D) graphics". Shouldn't it be enough to check only "printerGraphics", and if not, why not?

valerybokov:
> Why check both? I assume with "cast" you mean "printerGraphics = (Graphics2D) graphics". Shouldn't it be enough to check only "printerGraphics", and if not, why not?

I hope I understood you correctly. Line 298 (new version): the graphics2D reference doesn't always refer to the same object as the printerGraphics reference. Therefore, we only dispose the object if a new graphics object has been created.

THausherr:
I'm not asking about the dispose, I'm asking about the line "if (printerGraphics != null && callerTransform != null)". printerGraphics is no longer null as soon as the assignment on line 221 is successful. So why check whether callerTransform  isn't null? It will always be non null if the casted assignment was successful.

valerybokov:
> I'm not asking about the dispose, I'm asking about the line "if (printerGraphics != null && callerTransform != null)". printerGraphics is no longer null as soon as the assignment on line 221 is successful. So why check whether callerTransform isn't null? It will always be non null if the casted assignment was successful.

Let's say this is a standard check, but it's really unlikely. Here are three cases:

1. If the graphics variable isn't a `Graphics2D`, a `ClassCastException` will occur.
2. There's an OOM error on line 223 (`printerGraphics.getTransform()`).
3. The graphics parameter is null. Then there will be a null pointer exception.

For all these cases, the result is the same: the finality block will be triggered, and the `printerGraphics` and `callerTransform` variables will be null.

I can simplify the check if you want.

THausherr:
Re (2) you're really going to extremes. If there is an OOM we're doomed anyway, it won't matter that we restore printerGraphics settings. Re (1) I've never heard of the cast to Graphics2D failing. A google search finds nothing, ChatGPT says it's historical.

xzel23:
Isn't the standard in Swing to simply work on a temporary context obtained by calling context.create() if you don't want changes to transformation etc to persist? When done, simply forget about it. No need to try to reset the original context to its former state. Should also be more performant.

THausherr:
Yes, there is an SO answer about this
https://stackoverflow.com/questions/901046/
and ChatGPT agrees, and suggests this
```
Graphics2D g2 = (Graphics2D) graphics.create();
try {
    // modify freely
} finally {
    g2.dispose();
}
```

xzel23:
@valerybokov I think you should really follow the graphics.create() suggestion. It's the established standard way to assert the value graphics is not changed and will perform better and IMHO it's also more readable.

THausherr:
I also support that. I didn't know it before, but I almost never do printing except here.

PR #436, opened 2026-04-05
GlyphLayoutProcessor for correct glyph layout and support of DIN 91379 (Issue PDFBOX-4951)

vk-github18:
GlyphLayoutProcessor for correct glyph layout and support of DIN 91379
using AWT classes for glyph layout based on HarfBuzz library.

Can be switched on for individual ContentStream or Acroform.
Does not change processing for other documents.

Works in single threading and in multi threading environments, you can use different instances of GlyphLayoutProcessor in different threads, but each instance of GlyphLayoutProcessor can only be used in one thread.

See DoGlyphLayoutDin91379.java, DoGlyphLayoutDin91379Form.java, DoGlyphLayoutLigaturesAndKerning.java, DoGlyphLayoutBidi.java, DoGlyphLayoutBengali.java, DoGlyphLayoutMissingGlyphs.java, and DoGlyphLayoutSMP.java for usage.

Example document showing correct rendering of all letters of DIN 91379: [DoGlyphLayoutDin91379.pdf](https://github.com/user-attachments/files/26579035/DoGlyphLayoutDin91379.pdf)

Example form: [DoGlyphLayoutDin91379Form.pdf](https://github.com/user-attachments/files/26579036/DoGlyphLayoutDin91379Form.pdf)

Example for ligatures and kerning: 
[DoGlyphLayoutLigaturesAndKerning.pdf](https://github.com/user-attachments/files/28215432/DoGlyphLayoutLigaturesAndKerning.pdf)

Example for bidirectional text:
[DoGlyphLayoutBidi.pdf](https://github.com/user-attachments/files/28215439/DoGlyphLayoutBidi.pdf)

Examples for Supplementary Multilingual Plane:
[DoGlyphLayoutSMP.pdf](https://github.com/user-attachments/files/28341918/DoGlyphLayoutSMP.pdf)

Error message for missing glyph:
<pre>
Exception in thread "main" java.lang.IllegalArgumentException: Missing glyph in font 'Lohit Bengali' for the character 'A', codePoint: 65 (U+0041).
	at org.apache.pdfbox.pdmodel.GlyphLayoutProcessor.checkMissingGlyphs(GlyphLayoutProcessor.java:91)
	at org.apache.pdfbox.pdmodel.GlyphLayoutProcessor.computeGlyphVector(GlyphLayoutProcessor.java:180)
	at org.apache.pdfbox.pdmodel.GlyphLayoutProcessor.showText(GlyphLayoutProcessor.java:210)
[...]

PR #450, opened 2026-05-18
Example for glyph positioning

vk-github18:
Example for glyph positioning
no changes of PDFBox classes.

Example: [ShowCharactersOfDin91379.pdf](https://github.com/user-attachments/files/27972839/ShowCharactersOfDin91379.pdf)

See https://issues.apache.org/jira/browse/PDFBOX-4951


vk-github18:
https://github.com/apache/pdfbox/pull/436 is integrated now

PR #452, opened 2026-05-23
set capacity for list in COSWriterCompressionPool.addStructure method

valerybokov:


lehmi:
The size of the given list is not the size of the new one. Depending on the contained objects it might be total different, smaller, bigger or maybe the same size.

PR #462, opened 2026-06-02
release resources in COSOutputStream

valerybokov:


msahyoun:
Since filters is passed in from outside, COSOutputStream doesn't own it — won't clear() here risk surprising the caller?

THausherr:
Yeah why not just set it to null.

valerybokov:
> Since filters is passed in from outside, COSOutputStream doesn't own it — won't clear() here risk surprising the caller?

Yeah, my bad.

valerybokov:
> Yeah why not just set it to null.

This is good idea, but it is final. Can we set it as not final?

msahyoun:
given how short-lived COSOutputStream is I don't see the benefit in removing final from filters just to ensure nulling filters. It will be GC'd anyway as soon as the COSOutputStream reference is no longer held

PR #463, opened 2026-06-02
release resources in COSInputStream

valerybokov:


msahyoun:
nit: the brace style of the new close() method uses same-line opening braces, while the rest of COSInputStream uses Allman style per PDFBox standard

THausherr:
Copilot suggests this could be refactored so that it stores only the last element of decodeResults and that makes sense. And if we null that one, then we should mention in the javadoc that getDecodeResult() is not to be called after close().

valerybokov:
> Copilot suggests this could be refactored so that it stores only the last element of decodeResults and that makes sense. And if we null that one, then we should mention in the javadoc that getDecodeResult() is not to be called after close().

This is bad idea to use the stream after its closing. Any method can throw IllegalOperationException (the stream is closed). IMHO, you shouldn't write something like "this method can/cannot be called when the stream is closed". Any method not recommended to call when the stream is closed.

THausherr:
I've refactored the whole thing so there's less memory usage. We don't need to clean up everything, normally ideally the gc should do this.

PR #464, opened 2026-06-02
release resources in CCITTFaxEncoderStream

valerybokov:


msahyoun:
// Release large data structures immediately so the GC can reclaim memory
        // even if callers hold a reference to this COSDocument instance.
        
         Is this a copy and paste error? This is not COSDocument

PR #465, opened 2026-06-02
release resources in PDAbstractContentStream

valerybokov:


msahyoun:
// Release large data structures immediately so the GC can reclaim memory
        // even if callers hold a reference to this COSDocument instance.
        
        Is this a copy and paste error? This is not COSDocument

THausherr:
I ran my rendering tests with a trace on these structures. Most of the time they're empty or have one element. Which isn't surprising considering that it's a stack. That there is 1 element is because it isn't required to include everything in q...Q operators.

PR #466, opened 2026-06-05
release resources in CCITTFaxEncoderStream

valerybokov:


THausherr:
After committing I noticed you didn't set inputBuffer to null to, was there a reason not to?

valerybokov:
> After committing I noticed you didn't set inputBuffer to null to, was there a reason not to?

I don't see any variable named formatBuffer here. But I did notice a small error: the fields must be nulled before calling the close method, because an IOException might occur. In that case, the fields won't be nulled.

valerybokov:
Now I understand, I checked another class. The inputBuffer field is final. That is why I not cleared it.

THausherr:
inputBuffer, not formatBuffer. 
Re "the fields won't be nulled" - that means the old behavior.

valerybokov:
> inputBuffer, not formatBuffer. Re "the fields won't be nulled" - that means the old behavior.

image

THausherr:
ok, I changed all that. However after that I realized this is one of the files I usually avoid changing because this code is from the twelvemonkeys project and I like to keep it as it is so that I can see easily what changes they do. This closing of resources isn't that important, ideally the gc will do it.

valerybokov:
> ok, I changed all that. However after that I realized this is one of the files I usually avoid changing because this code is from the twelvemonkeys project and I like to keep it as it is so that I can see easily what changes they do. This closing of resources isn't that important, ideally the gc will do it.

The GC will do it when more memory needed or when timer inside it will call clearing algorithm. But the less number of the links (more nullified links) the faster memory clearing and the less likely to run out of memory.

PR #468, opened 2026-06-05
fix-comment-for-IOUtils.applyOwnerOnlyPermissions method

valerybokov:


msahyoun:
Since LinkOption is only referenced in the Javadoc and not used in actual code, I'd suggest dropping the import and using the fully qualified name in the @see tag instead:
``` java
@see Files#getFileAttributeView(Path, Class, java.nio.file.LinkOption[])
```
This keeps the import list clean and ensures it only reflects classes actually used in the code.

PR #483, opened 2026-07-19
fix log messages format in PNGConverter

valerybokov:
There were multiple warnings in this class. The log message format was incorrect. Reason: We can't use %d in this case.


THausherr:
Thanks, I believe that I tested months ago that this worked, but this doesn't matter because I read that {} is faster anyway.

PR #484, opened 2026-07-19
Fix CCITTFactory.readlong

valerybokov:
## Summary

`CCITTFactory.readlong()` reads a TIFF `LONG` field (an unsigned 32-bit value per the TIFF spec) by assembling four bytes with an `int`, then returning that `int`. At the call site (`long address = readlong(...)`), the `int` is widened to `long`, which in Java **sign-extends**. If the top bit of the 32-bit value is set (i.e. the raw value is `>= 0x80000000`), the widened `long` becomes a large *negative* number instead of the correct unsigned value — corrupting IFD offsets used directly in `reader.seek(address)`.

In practice this is latent rather than commonly hit: classic TIFF (non-BigTIFF) is capped near 4GB by its own 32-bit offset fields, and CCITT G3/G4-compressed files in particular are typically KB-to-low-MB in size, so authentic offsets essentially never cross that boundary. The realistic risk is a malformed/corrupted/adversarial TIFF whose 4-byte offset or count field has the high bit set — `readlong` trusts those bytes without validating against the actual file size, so such input silently produces a corrupted negative seek target rather than a controlled error.

## Fix

- fix warning for `readlong` method
- `readlong` now returns `long` and masks the assembled value with `& 0xFFFFFFFFL`, so it's always treated as unsigned when read.
- `address` (already declared `long`) now receives the correct unsigned value automatically.
- `count` and `val`, which are intentionally kept as `int` (small tag values used with `params.setInt(...)` / buffer sizes), now narrow explicitly via `(int)` casts — same bit pattern as before, so their behavior is unchanged.

## Test plan

Added `CCITTFactoryTest#testReadLongIsUnsigned`, which invokes the private `readlong` via reflection against crafted byte sequences with the high bit set (`0xFFFFFFFF` and `0x80000000`, in both byte orders) and asserts the returned `long` is the correct non-negative unsigned value.

- Verified the test fails against the pre-fix `int`-returning implementation (e.g. expected `4294967295`, got `-1`) and passes against the fix.
- Existing `CCITTFactoryTest` suite (small real-world CCITT G3/G4 fixtures) continues to pass — those files never exercised this path since their offsets are well under `0x80000000`.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CCITTFactory.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/CCITTFactory.java
@@ -303,7 +303,7 @@ public final class CCITTFactory
             {
                 int tag = readshort(endianess, reader);
                 int type = readshort(endianess, reader);
-                int count = readlong(endianess, reader);
+                int count = (int) readlong(endianess, reader);
                 int val;
                 // Note that when the type is shorter than 4 bytes, the rest can be garbage
                 // and must be ignored. E.g. short (2 bytes) from "01 00 38 32" (little endian)
@@ -322,7 +322,7 @@ public final class CCITTFactory
                         reader.read();
                         break;
                     default: // long and other types
-                        val = readlong(endianess, reader);
+                        val = (int) readlong(endianess, reader);
                         break;
                 }
                 switch (tag)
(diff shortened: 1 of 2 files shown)


THausherr:
Hi, assuming this is AI supported, please read this: https://www.apache.org/legal/generative-tooling.html and please disclose the AI name in contributions. Also please change the PR so that the method to be tested is package private so that you don't have to use reflection.

valerybokov:
https://github.com/apache/pdfbox/pull/488

PR #492, opened 2026-07-27
PDFBOX-6077: Add regression tests for stencil-mask pattern fills

valerybokov:
Summary

Adds regression tests for the two bugs fixed on the fix-image-masks branch (from which this branch is built): a stencil-mask image filled with a pattern, where either the pattern itself has gaps (e.g. between tiles) or a soft mask is applied to the pattern (https://github.com/apache/pdfbox/pull/491).

What's added

- pom.xml — two download-maven-plugin entries fetching the JIRA-attached test PDFs into target/pdfs, following the project's existing convention for issue-specific fixtures (SHA-512-pinned download, same pattern as the neighboring PDFBOX-4831 entry):
  - PDFBOX-6077-example.pdf (the originally reported file)
  - PDFBOX-5842-reduced.pdf (the related file a naive fix for the first bug alone regresses)
- TestQuality.java — two new tests:
  - testPDFBox6077: renders example.pdf and asserts a pixel in a gap between the tiling pattern's tiles is white (0xFFFFFFFF), not opaque black. Confirmed this fails on pre-fix code (expected: <-1> but was: <-16777216>).
  - testPDFBox5842: renders PDFBOX-5842-reduced.pdf and asserts a pixel inside the soft-masked pattern's map-marker icon isn't blank white — guards against the soft-mask fix regressing (which an earlier, discarded fix attempt did).

This is a rough draft of the tests, as the POM file will likely need to be modified.

--------------------------
Per https://www.apache.org/legal/generative-tooling.html: portions of this PR were produced with assistance from Claude Code (Anthropic), based on a bug from PDFBOX Issue Tracker (6077).

I've reviewed the generated code and test, verified the test fails against the pre-fix implementation and passes against the fix, and confirm to the best of my knowledge that the output does not include any third-party copyrighted material and is compatible with the Apache License 2.0.


THausherr:
What about this one? It's still a draft and has non-test code. We may be able to use the test (I haven't tried it) but it's possible that the test fails on our CI because different machines and different jdks produce different renderings. But it's worth a try.

valerybokov:
> What about this one? It's still a draft and has non-test code. We may be able to use the test (I haven't tried it) but it's possible that the test fails on your CI because different machines and different jdks produce different renderings. But it's worth a try.

Yes, I remember about this. I think, I will refresh this code this week.

PR #496, opened 2026-08-13
fix JPEGFactoryTest.computeMeanAbsDiffPerPixel method

valerybokov:


THausherr:
Ooops. Thanks. This happened because while working on this I initially compared an int.

PR #497, opened 2026-08-17
optimize JPEGFactory methods

valerybokov:
The current version of the JPEGFactory.createFromByteArray method always creates a ByteArrayInputStream instance to read the image dimensions, and then creates a PDImageXObject instance. If we use the JPEGFactory.createFromStream method, we have a stream instance and create a new array. This is inefficient. We can work with streams and avoid duplicating memory (the InputStream from createFromStream can also be a ByteArrayInputStream).
I thought the InputStream.markSupported method should be used, and I found this information. The BufferedInputStream.reset() method can throw an exception only in a pathological extreme case: if, after mark(), more bytes were read than fit in the Java array (~2 GB). This isn't a drawback specific to BufferedInputStream—it's a fundamental limitation on storing arbitrary, rewindable data in memory, and it affects all approaches, including the original code before the refactoring (stream.readAllBytes()), which also resulted in an error (out of memory or exceeding array size limits) when handling multi-gigabyte input data. Therefore, BufferedInputStream is no worse than the alternative—it's just as good for realistic input data, with the same theoretical limit as everything else.
If this isn't acceptable, I can rewrite it using the markSupported method.
An additional benefit: the PDImageXObject.createFromFileByContent method now uses BufferedInputStream via the JPEGFactory.createFromStream method.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/JPEGFactory.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/JPEGFactory.java
@@ -81,7 +81,7 @@ public final class JPEGFactory
     public static PDImageXObject createFromStream(PDDocument document, InputStream stream)
             throws IOException
     {
-        if (!(stream instanceof ByteArrayInputStream) && !(stream instanceof BufferedInputStream))
+        if (!stream.markSupported())
         {
             stream = new BufferedInputStream(stream);
         }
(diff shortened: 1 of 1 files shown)


pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/image/JPEGFactory.java
```diff
@@ -80,25 +81,13 @@ private JPEGFactory()
...
-     * @return a new Image XObject
-     *
-     * @throws IOException if the input stream cannot be read
-     */
-    public static PDImageXObject createFromByteArray(PDDocument document, byte[] byteArray)
-            throws IOException
-    {
-        // copy stream
-        ByteArrayInputStream byteStream = new ByteArrayInputStream(byteArray);
+        if (!(stream instanceof ByteArrayInputStream) && !(stream instanceof BufferedInputStream))
```
THausherr:
```suggestion
        if (!stream.markSupported())
```

valerybokov:
@THausherr , but there is one small thing. The method markSupported() can return true but the stream will not be buffered.

THausherr:
I agree, e.g. for a FileInputStream. But the result of the last change is that we don't do more buffering than we did before.

PR #498, opened 2026-08-21
Refactor PfbParser.parsePfb

valerybokov:
Use the stream as the main source for the PfbParser method.

### DIFF
--- a/fontbox/src/main/java/org/apache/fontbox/pfb/PfbParser.java
+++ b/fontbox/src/main/java/org/apache/fontbox/pfb/PfbParser.java
@@ -17,12 +17,12 @@
 package org.apache.fontbox.pfb;
 
 import java.io.ByteArrayInputStream;
+import java.io.BufferedInputStream;
 import java.io.EOFException;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


THausherr:
There's one problem: `available()` is unreliable (it happened to us elsewhere). That part of the code needs the actual length.
Btw this part of the isn't used much, this is about PFB fonts which are only used by very old applications.

valerybokov:
> There's one problem: `available()` is unreliable (it happened to us elsewhere). That part of the code needs the actual length. Btw this part of the isn't used much, this is about PFB fonts which are only used by very old applications.

Thanks for telling me, but I'd rather complete this pull request.

The latest commit description:

We read header of the file twice to calculate header size. To start read again we used the mark()/reset() methods.

The commit [0c9829a](https://github.com/apache/pdfbox/pull/498/commits/0c9829aaf6cc5f8caef1cd03b0b0cb33c3572386) avoids this double job.
Segment reads now use readNBytes(size) instead of new byte[size] + read(), so an oversized/bogus segment size can never allocate more than the stream actually contains — this replaces the old size > availableSize / total > availableSize checks (which relied on available(), an unreliable size estimate). Since readNBytes already bounds each segment to real data, total can no longer exceed what's actually in the stream, making those checks redundant.

THausherr:
Thanks, this is really nice now.

PR #501, opened 2026-08-27
do not close the stream inside CCITTFactory.extractFromTiff method

valerybokov:
This code doesn't throw an IOException or other error, but the code structure is incorrect. It is fragile. We write data to the output stream in the extractFromTiff method, close it, and then read the data from outside in the createFromRandomAccessImpl method. But the stream is closed. We can do it only because this is ByteArrayOutputStream. If we change the stream type then there will be an exception.


THausherr:
Thank you. It turns out I introduced this in 2014 (before that the close was needed). I didn't use the PR, I remove the close manually, hopefully that is the same.

PR #504, opened 2026-08-29
delete always the same offset argument in RC4Cipher.write method

valerybokov:
This is the private method and the argument is always zero.


THausherr:
Thanks, I'll remove `j` as well.

PR #505, opened 2026-08-29
use SecurityHandler.encryptDataRC4(byte[], byte[], OutputStream) method

valerybokov:
We can call another encryptDataRC4 method to avoid allocating memory for ByteArrayInputStream.

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/encryption/StandardSecurityHandler.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/encryption/StandardSecurityHandler.java
@@ -16,7 +16,6 @@
  */
 package org.apache.pdfbox.pdmodel.encryption;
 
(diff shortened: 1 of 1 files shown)


THausherr:
Thank you, nice find. This will allow me to deprecate / remove that other method.
https://issues.apache.org/jira/browse/PDFBOX-6247

valerybokov:
> Thank you, nice find. This will allow me to deprecate / remove that other method.

Okay. There is no reason to use it.
As far as I understand, encryption class methods typically use streams rather than arrays. However, this is more my observation than definitive knowledge.

THausherr:
Sadly it was too good to be true. Some weird error in my own thought.

valerybokov:
> Sadly it was too good to be true. Some weird error in my own thought.

Do you mean it still used some were? Yeah. Maybe you will do it later with changed code.

PR #507, opened 2026-08-29
fix swapped error messages in CCITTFactory.extractFromTiff method

valerybokov:
See: TIFFExtension.GROUP3OPT_UNCOMPRESSED and TIFFExtension.GROUP3OPT_FILLBITS.
I would prefer to use these constants there.


THausherr:
Sorry I didn't see that you had noticed the mixup as well.

valerybokov:
> Sorry I didn't see that you had noticed the mixup as well.

Okay, we fixed it somehow.

PR #512, opened 2026-08-30
fix comments in CRLVerifier type

valerybokov:


THausherr:
thanks but not needed

PR #516, opened 2026-09-01
fix comments for COSName type

valerybokov:


THausherr:
No this is correct as it is: https://en.wikipedia.org/wiki/Canonicalization

PR #519, opened 2026-09-04
refactor the FlateFilterDecoderStream.mark and reset methods

valerybokov:
The FlateFilterDecoderStream.mark and reset methods may be not synchronized because they do not use any fields or methods.


THausherr:
Reverted because of Sonar complaint. It doesn't matter anyway, that part isn't called.
grafik

valerybokov:
> Reverted because of Sonar complaint. It doesn't matter anyway, that part isn't called. grafik

Why can't you disable some Sonar warning? Stupid SonarCloud.

PR #520, opened 2026-09-04
add the TestQuality.testPDFBox6232

valerybokov:
- Added testPDFBox6232() to TestQuality.java — confirmed it fails on the original code (blue channel 0x52=82, matching the ticket's exact "~80" figure) and passes with the fix.
- Added the reproducer PDF to the rendering smoke-test resources; the full TestRendering suite (11 files) passes.
-------------------------------------------------------
- Per https://www.apache.org/legal/generative-tooling.html: portions of this PR were

### DIFF
--- a/pdfbox/pom.xml
+++ b/pdfbox/pom.xml
@@ -1033,6 +1033,19 @@
                             <sha512>0d7160a4af04bf2f785bf4155f69876da4b7ff7a196e7eda1eb904c02c35aa03c31041d9339fda90322cb58bde16b87bae8460c617bfe5b8dc0670ddab102f62</sha512>
                         </configuration>
                     </execution>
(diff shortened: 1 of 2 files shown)


valerybokov:
Hi, @THausherr! This is the changed code to use smaller file what you attached into IssueTracker.

[PDFBOX-6232-use-reduced-file.patch](https://github.com/user-attachments/files/31863493/PDFBOX-6232-use-reduced-file.patch)

THausherr:
I wouldn't use that file, because that one isn't rendered correctly with the change. It is unclear whether this is because of the change or because of a different reason (I deleted a lot in that file).

THausherr:
Memo for me: this is the issue where we're waiting for feedback.

PR #525, opened 2026-09-06
Fix pdfbox 5876

valerybokov:
Fix: bound that metadata-only read to a fixed, sane subsampling (8 — matching the max PageDrawer.getSubsampling() ever requests anyway) instead of either "no limit" or MAX_VALUE. reader.getWidth(0)/getHeight(0) are unaffected by subsampling, so dimensions stay exact; only the resolution of an embedded soft mask (SMaskInData, rare) is affected, and it's already rescaled elsewhere.

- Added TestQuality.testPDFBox5876() (spawns a -Xmx600m subprocess, since heap can't be changed mid-JVM) — verified it fails with the exact original stack trace when the fix is reverted, and passes with it applied.

Changes: pdfbox/pom.xml (new test-fixture download entry), PDImageXObject.java (the fix), TestQuality.java + new JPXLowMemoryRenderMain.java (the regression test).

-------------------------------------

### DIFF
--- a/pdfbox/pom.xml
+++ b/pdfbox/pom.xml
@@ -1046,6 +1046,19 @@
                             <sha512>cd6d92c643108cdcf2b4c6a4ae4cc5a15d848e3335f6c8a135a5438d691eb982d11cfd8f26d11839321e06a364126d9b0c9732b4a19785821b844fe0ec94d525</sha512>
                         </configuration>
                     </execution>
+                    <execution>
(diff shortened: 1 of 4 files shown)


THausherr:
I haven't time to test this but this looks like a great idea.

valerybokov:
> I haven't time to test this but this looks like a great idea.

You don't have to test it today. Plan to do it in a few days.

PR #528, opened 2026-09-08
Increase chunk-size cap in Filter.decode() for large streams

valerybokov:
The buffer chunk-size estimate in the static decode() collapsed to the
flat 4KB default for any encoded stream >= 1KB, forcing thousands of
separate 4KB chunk allocations in RandomAccessReadWriteBuffer when
decoding large images/fonts. Cap the estimate at 1MB instead of
falling back to 4KB, cutting the number of chunk allocations for large
streams while still bounding worst-case preallocation

### DIFF
--- a/pdfbox/src/main/java/org/apache/pdfbox/filter/Filter.java
+++ b/pdfbox/src/main/java/org/apache/pdfbox/filter/Filter.java
@@ -279,14 +279,20 @@ public abstract class Filter
                 length = randomAccessWriteBuffer.length();
             }
             // we don't know the size of the decoded stream, just estimate a 4 times bigger size than the encoded stream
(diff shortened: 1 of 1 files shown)


valerybokov:
Another question. It looks like the issue https://issues.apache.org/jira/browse/PDFBOX-5491 is okay now. Some code fixed a problem.

THausherr:
[PDFBOX-5491](https://issues.apache.org/jira/browse/PDFBOX-5491) still happens for me.

valerybokov:
> [PDFBOX-5491](https://issues.apache.org/jira/browse/PDFBOX-5491) still happens for me.

Okay. I'll check one more time.

PR #529, opened 2026-09-08
Fix pdfbox 5250

valerybokov:
In PDFStreamEngine.processTransparencyGroup(), initialMatrix was captured from the CTM before the transparency group's own /Matrix entry got concatenated into it — the opposite order used by processStream() (the code path for ordinary, non-group forms). Per spec, a pattern's matrix maps pattern space to "the default coordinate system of the pattern's parent content stream," which must include that stream's own /Matrix. So any shading/tiling pattern used inside a transparency group that itself has a non-trivial /Matrix (like the reported case's 0.017 scale) gets positioned using the wrong reference matrix — off by exactly that scale factor.


THausherr:
Thanks, this fixes PDFBOX-4077 as well, and a file from the poppler project. I'll run the actual fix now and the test later this week when I have more time.

